PR #39, opened 2015-01-25
Fix broken links in documentation that go to different sections.

gching:
Fix links that are suppose to direct the user to other sections instead of directing them to a 404. Small simple fix :).


pozadi:
Thanks, found one of those lately myself, but didn't notice there are more.

PR #47, opened 2015-02-10
make the classes Pool and Bus publicly available

mhelvens:
Kefir has functions like [`Kefir.pool`](https://github.com/pozadi/kefir/blob/c7b60ae9550a75e5d565c5f8e6f34ab7c8cfb998/src/multiple-sources.js#L197) and [`Kefir.bus`](https://github.com/pozadi/kefir/blob/c7b60ae9550a75e5d565c5f8e6f34ab7c8cfb998/src/multiple-sources.js#L239). What they return are actually new instances of the classes [`Pool`](https://github.com/pozadi/kefir/blob/c7b60ae9550a75e5d565c5f8e6f34ab7c8cfb998/src/multiple-sources.js#L178) and [`Bus`](https://github.com/pozadi/kefir/blob/c7b60ae9550a75e5d565c5f8e6f34ab7c8cfb998/src/multiple-sources.js#L207). But users of Kefir have no access to these constructors.

Was there a reason for hiding them? Even if you find these 'factory functions' more elegant as a way to instantiate these types of streams, I would still prefer to have access to the constructors for purposes of documentation, type-checking, and so on.

Note that these are not just stream implementations. They have an extended public interface. I do not, for example, propose making the [`FlatMap`](https://github.com/pozadi/kefir/blob/c7b60ae9550a75e5d565c5f8e6f34ab7c8cfb998/src/multiple-sources.js#L249) constructor public.

### DIFF
--- a/src/multiple-sources.js
+++ b/src/multiple-sources.js
@@ -178,6 +178,7 @@ Observable.prototype.concat = function(other) {
 function Pool() {
   _AbstractPool.call(this);
 }
+Kefir.Pool = Pool;
 
 inherit(Pool, _AbstractPool, {
 
@@ -207,6 +208,7 @@ Kefir.pool = function() {
 function Bus() {
   _AbstractPool.call(this);
 }
+Kefir.Bus = Bus;
 
 inherit(Bus, _AbstractPool, {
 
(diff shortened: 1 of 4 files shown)


pozadi:
That makes sense. Wasn't sure to make them public because then we should make public `FlatMap` etc., but you made good point that this classes have their own methods and this is a good criterion by which we can decide whether make them public or not.

PR #52, opened 2015-02-14
fixed bower paths after update its components

saitodisse:


pozadi:
I don't use bower for installing packages directly, those are installed via `grunt bower` instead. I think everything works ok now, we don't need any fix for this, but thanks!

PR #61, opened 2015-02-23
Minor corrections and stuff

raimohanska:
Briefly checked the comparison doc. This is good stuff. Big thums up!

### DIFF
--- a/bacon-vs-kefir-api.md
+++ b/bacon-vs-kefir-api.md
@@ -8,7 +8,7 @@
 | ----- | ----- | -------- |
 | `$.fn.asEventStream(eventName, [selector], [eventTransformer])` | `$.fn.asKefirStream(eventName, [selector], [eventTransformer])` | In Kefir this method available not in core but in separate lib [kefir-jquery](https://github.com/pozadi/kefir-jquery) |
 | `Bacon.fromPromise(promise, [abort])` | `Kefir.fromPromise(promise)` | No `abort` option in Kefir, and the result is a Property unlike Stream in Bacon |
-| `Bacon.fromEventTarget(target, eventName, [eventTransformer])` | `Kefir.fromEvent(target, eventName, [eventTransformer])` |  |
+| `Bacon.fromEvent(target, eventName, [eventTransformer])` | `Kefir.fromEvent(target, eventName, [eventTransformer])` |  |
 | `Bacon.fromCallback(f, [args...])` | `Kefir.fromCallback(fn)` | No `args` argument in Kefir |
 | `Bacon.fromCallback(object, methodName, [args...])` | No alt. |  |
 | `Bacon.fromNodeCallback(f, [args...])` | `Kefir.fromNodeCallback(fn)` | No `args` argument in Kefir |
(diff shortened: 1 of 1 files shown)


pozadi:
Thanks! Was going to ask you for corrections :) 
Still a bit confused though, will add comments to the diff.

bacon-vs-kefir-api.md
```diff
@@ -8,7 +8,7 @@
 | ----- | ----- | -------- |
 | `$.fn.asEventStream(eventName, [selector], [eventTransformer])` | `$.fn.asKefirStream(eventName, [selector], [eventTransformer])` | In Kefir this method available not in core but in separate lib [kefir-jquery](https://github.com/pozadi/kefir-jquery) |
 | `Bacon.fromPromise(promise, [abort])` | `Kefir.fromPromise(promise)` | No `abort` option in Kefir, and the result is a Property unlike Stream in Bacon |
-| `Bacon.fromEventTarget(target, eventName, [eventTransformer])` | `Kefir.fromEvent(target, eventName, [eventTransformer])` |  |
+| `Bacon.fromEvent(target, eventName, [eventTransformer])` | `Kefir.fromEvent(target, eventName, [eventTransformer])` |  |
```
pozadi: This was from Bacon docs :)

bacon-vs-kefir-api.md
```diff
@@ -155,11 +155,11 @@
 | ----- | ----- | -------- |
 | `stream.map(property)` | Use `.sampledBy()` |  |
 | `obs.filter(property)` | `obs.filterBy(obs)` |  |
-| `obs.takeWhile(property)` | `obs.takeWhileBy(obs)` | Only property supported in Bacon? |
+| `obs.takeWhile(property)` | `obs.takeWhileBy(obs)` | Bacon supports property or predicate, Kefir supports any observable |
```
pozadi: There is versions of these methods with predicate in the "Modify an observable" section. I think we'd better add something like "see also `takeWhile(fn)` above", to not confuse anybody that Kefir doesn't have version with predicate.

bacon-vs-kefir-api.md
```diff
@@ -155,11 +155,11 @@
...
 | `obs.combine(obs2, fn)` | `obs.combine(obs2, [fn])` | In Kefir `fn` optional |
-| `stream.skipWhile(property)` | `obs.skipWhileBy(otherObs)` | Only property supported in Bacon? |
-| `stream.skipUntil(stream2)` | `obs.skipUntilBy(otherObs)` | Only stream supported in Bacon? |
-| `obs.takeUntil(stream) ` | `obs.takeUntilBy(otherObs)` | Only stream supported in Bacon? |
+| `stream.skipWhile(property)` | `obs.skipWhileBy(otherObs)` | Bacon supports property or predicate, Kefir supports any observable |
+| `stream.skipUntil(stream2)` | `obs.skipUntilBy(otherObs)` |  |
+| `obs.takeUntil(stream) ` | `obs.takeUntilBy(otherObs)` |  |
```
pozadi: Does that mean that `.skipUntil` and `.takeUntil` supports any observable in both positions? We should fix the signatures then too. Of if signatures correct, there should be a note.

raimohanska: Oh indeed in Bacon it is `stream.take/skipWhile(obs)`... So Property supported in second position only.

bacon-vs-kefir-api.md
raimohanska: True that!

bacon-vs-kefir-api.md
raimohanska: Well yeah. That's the old one. Who reads docs anyway? :) I'll update...

pozadi:
Ok, merging this, will make changes based on comments after merge. Thanks!

PR #86, opened 2015-04-11
Include dates of release in changelog

Merri:
Adds the dates with no special formatting.


pozadi:
I think it would be better if dates were in titles like in https://github.com/facebook/react/blob/master/CHANGELOG.md

pozadi:
Will fix formatting by myself. Thanks!

PR #106, opened 2015-05-24
Fix scan benchmark

paldepind:
Thank you for having created these benchmarks. They've been very helpful in improving the performance of [Flyd](https://github.com/paldepind/flyd).

Would you be interested in another PR adding Flyd to a few of the benchmark cases (don't be afraid, Kefir is fastest)?


rpominov:
Thanks for the fix.

> Would you be interested in another PR adding Flyd to a few of the benchmark cases (don't be afraid, Kefir is fastest)?

That would be cool!

rpominov:
Glad my tests helped, btw :)

PR #107, opened 2015-05-28
Add Flyd to several of the benchmarks

paldepind:
This PR adds [Flyd](https://github.com/paldepind/flyd) to a bunch of the benchmarks.


rpominov:
Nice, thanks!

Just curious, why you have `map` in core, but `filter` as a separate module?

paldepind:
You're not the first to ask the question! It's a bit arbitrary but I try to keep as little as possible in the core.

I might put it in if enough people ask ;)

iofjuupasli:
@paldepind I like the idea of using transducers with streams. So adding filter or anything else that can be implemented with transducers seems as unreasonable duplicating. Stream lib should do one thing, and do it well (piping over observable). For synchronous transforms we already have transducers. It's well standardized, and works very good for its tasks.
I don't see the reason to duplicate same code in kefir, flyd, rxjs, bacon, etc. if it can be nicely modularized with one standard protocol. Maybe I miss something, or it's a matter of time.

Issue #128, opened 2015-08-14
BufferWhileBy should emit empty array if primary source finished first


shamansir:
`BufferWhileBy` should emit empty array if primary source finished (without any value) before the secondary one.

shamansir:
Could be related to #113

rpominov:
Fixed by #129

Issue #103, opened 2015-05-09
Polish the API


rpominov:
Version 3.0.0 probably won't happen very soon, but we can start collecting ideas for it already. In particular, I think the API can be made more solid and less hairy in some places.

I'm personally not very happy with following parts:

 - `.valuesToErrors/errorsToValues` (method names and returning value shape)
 - bunch of subscription methods `.on*/.off*`
 - `.endOnError` probably should be `.takeErrors(n)` to line up with `.take(n)`

Basic rules:

 - We want to make API smaller and simpler, not other way. But without sacrificing functionality.
 - Ideally there should be only one way to do each thing. 
 - We want methods for errors to line up with similar methods for values.

Example of such a change (already done): add `.last()`, but remove `.reduce()`, so one can do `.scan().last()` instead of reduce, but also can do just `.last()`.

Any ideas?

-----------
Related: #71, http://www.youtube.com/watch?v=4anAwXYqLG8

pozadi:
Have an idea how we could fix mess with subscription methods:
 1. Change [emitter](http://pozadi.github.io/kefir/#emitter-object) interface from `{emit, error, end, emitValue}` to `{value, error, end, event}`
 2. Add `Observable.prototype.observe()` that accepts an object with same shape as `emitter`, and returns a `dispose` function.
 3. If `observe()` called with a function as an argument, it automatically converted to `{value: passedFn}`

| Before | After  |
| ----- | ----- | 
| `obs.onValue(fn);`<br/>`obs.offValue(fn);` | `let dispose = obs.observe(fn);`<br/>`dispose();` |
| `obs.onError(fn);`<br/>`obs.offError(fn);` | `let dispose = obs.observe({error: fn});`<br/>`dispose();` |
| `obs.onValue(fn1).onError(fn2);` | `let dispose = obs.observe({value: fn1, error: fn2});` |

Full disclosure: idea of same interface for observer and emitter borrowed from RxJS, returning function to unsubscribe — from Bacon, the method name — from [Most](https://github.com/cujojs/most)

pozadi:
We should also look to where https://github.com/jhusain/asyncgenerator moving. As I understand current proposition for subscribing in there looks like this:

```js
stream[@@observer](generator);
```

In other words:

```js
stream[@@observer]({
  next(x) {/* you can return `{done: true}` from here to unsubscribe */ ...},
  throw(err) {...},
  return(x) {...}
});
```

[...]

PR #129, opened 2015-08-14
BufferWhileBy: emit empty array when primary source finished first (#128)

shamansir:
See #128

### DIFF
--- a/test/specs/buffer-while-by.coffee
+++ b/test/specs/buffer-while-by.coffee
@@ -20,11 +20,17 @@ describe 'bufferWhileBy', ->
       b = prop()
       expect(a.bufferWhileBy(b)).toActivate(a, b)
 
-    it 'should flush empty buffer and then end when primary ends', ->
-      expect(send(stream(), ['<end>']).bufferWhileBy(stream())).toEmit [{current: []}, '<end:current>']
+    it 'should end when primary ends (w/o {emitEmpty: true})', ->
+      expect(send(stream(), ['<end>']).bufferWhileBy(stream())).toEmit ['<end:current>']
       a = stream()
       b = stream()
-      expect(a.bufferWhileBy(b)).toEmit [[], '<end>'], -> send(a, ['<end>'])
+      expect(a.bufferWhileBy(b)).toEmit ['<end>'], -> send(a, ['<end>'])
+
+    it 'should flush empty buffer and then end when primary ends (w/ {emitEmpty: true})', ->
+      expect(send(stream(), ['<end>']).bufferWhileBy(stream(), {emitEmpty: true})).toEmit [{current: []}, '<end:current>']
+      a = stream()
+      b = stream()
+      expect(a.bufferWhileBy(b, {emitEmpty: true})).toEmit [[], '<end>'], -> send(a, ['<end>'])
 
     it 'should flush buffer on end', ->
       expect(send(prop(), [1, '<end>']).bufferWhileBy(stream())).toEmit [{current: [1]}, '<end:current>']
@@ -161,4 +167,3 @@ describe 'bufferWhileBy', ->
       a = send(prop(), [1])
       b = send(prop(), [true])
       expect(a.bufferWhileBy(b)).toEmit []
-
--- a/test/specs/buffer-while-by.coffee
+++ b/test/specs/buffer-while-by.coffee
@@ -32,6 +32,12 @@ describe 'bufferWhileBy', ->
       b = stream()
       expect(a.bufferWhileBy(b, {emitEmpty: true})).toEmit [[], '<end>'], -> send(a, ['<end>'])
 
+    it 'should flush empty buffer and then end when primary ends (w/ {flushOnChange: true, emitEmpty: true})', ->
+      expect(send(stream(), ['<end>']).bufferWhileBy(stream(), {flushOnChange: true, emitEmpty: true})).toEmit [{current: []}, '<end:current>']
+      a = stream()
+      b = stream()
+      expect(a.bufferWhileBy(b, {flushOnChange: true, emitEmpty: true})).toEmit [[], '<end>'], -> send(a, ['<end>'])
+
     it 'should flush buffer on end', ->
       expect(send(prop(), [1, '<end>']).bufferWhileBy(stream())).toEmit [{current: [1]}, '<end:current>']
       a = stream()
--- a/test/specs/buffer-while-by.coffee
+++ b/test/specs/buffer-while-by.coffee
@@ -32,11 +32,11 @@ describe 'bufferWhileBy', ->
       b = stream()
       expect(a.bufferWhileBy(b, {emitEmpty: true})).toEmit [[], '<end>'], -> send(a, ['<end>'])
 
-    it 'should flush empty buffer and then end when primary ends (w/ {flushOnChange: true, emitEmpty: true})', ->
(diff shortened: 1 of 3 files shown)


rpominov:
I actually thought of making this as an option this time, rather than introducing a breaking change. Should've done the same for `filterBy`, but haven't had that thought at the right time :(

Let's add option `emitEmpty` alongside with others (`flushOnEnd` & `flushOnChange `), `false` by default.
There will be an inconsistency with `filterBy`, that has that behavior by default, but we'll fix this in the future.

What do you think?

shamansir:
Yeah, that's probably better than breaking everything. Though these two, `bufferBy`/`bufferWhileBy`, will be most customizable among others :).

shamansir:
Done, I suppose, see new change :)

rpominov:
Also could you add a test case with `{flushEmpty: true, flushOnChange: true}`, when the controlling stream "chnages" before any value from main stream so the result stream emits `[]`?

rpominov:
The new test case is same as test with emitting `[]` on end case (which you've added earlier).

There is two cases now when the result stream can emit `[]`:
 - when the main stream ends,
 - when controlling stream changes from `true` to `false` (simply emits `false` actually).

We have a test case for the first one, but not for the second.

shamansir:
```
it 'should flush empty buffer and then end when secondary ends (w/ {flushOnChange: true, emitEmpty: true})', ->
      expect(stream().bufferWhileBy(send(stream(), ['<end>']), {flushOnChange: true, emitEmpty: true})).toEmit [{current: []}, '<end:current>']
      a = stream()
      b = stream()
      expect(a.bufferWhileBy(b, {flushOnChange: true, emitEmpty: true})).toEmit [[], '<end>'], -> send(b, ['<end>'])
```

Is this one right? It fails in my case, should I fix it?

shamansir:
(fails since emits just an empty array without `end` event)

shamansir:
With a lot of "intersecting" options, it's becomes a bit harder for one's mind to get what actually should happen :)

rpominov:
> Is this one right?

Not quite, its should be `send(b, [true, false])`

> fails since emits only empty array without end event

It's not necessarily have to end after emitting `[]` (see explanation below)

> With a lot of "intersecting" options, it's becomes a bit harder for one's mind to get what actually should happen :)

Yeah, sorry. I'll try to explain how it works. But perhaps you are right that we should simplify it somehow for easier understanding (maybe in v3.0 #103).

So here how it works. It buffers all the values from main stream, and may flush the buffer in 3 cases:

 - On a value from main stream, if latest value from controlling stream was `false`.
 - On a `false` value from controlling stream, if `{flushOnChange: true}`.
 - On end of main stream, unless `{flushOnEnd: false}`.

And normally in all those 3 cases we check if buffer not empty before emit. But if `{emitEmpty: true}`, we don't do that check. Also in the first case the buffer always has at least 1 value, because there just was a value from main stream, so `emitEmpty` does't matter here.

Also note that in the second case after emitting `[]`, the result stream doesn't end, why should it?

Hope that'll help :)

shamansir:
Yes, that's a great explanation, thank you! I think I got this right this time.

shamansir:
> On a value from main stream, if latest value from controlling stream was `false`
if `flushOnChange` is `false`, I suppose :)

shamansir:
I'll be able to fix it in the evening, I hope there's no rush :blush:

rpominov:
> if `flushOnChange` is false, I suppose :)

No, `flushOnChange` also doesn't matter here. On a value from main it puts the new value to the buffer and then immediately flushes the buffer.  


With `{flushOnChange: false}`:

```
foo:     ----1----2----3----4----5----6----7----8X
bar:     -----------f---------t---------fX

result:  --------------•----•--------------•----•X
               [1, 2, 3]  [4]      [5, 6, 7]  [8]
```

With `{flushOnChange: true}`:

```
foo:     ----1----2----3----4----5----6----7----8X
bar:     -----------f---------t---------fX

result:  -----------•--•----•-----------•--•----•X
               [1, 2] [3]  [4]     [5, 6] [7]  [8]
```

> I'll be able to fix it in the evening, I hope there's no rush :blush:

Of course, no problem :)

shamansir:
Hope now I got things right. So I send `true` and `false` to a secondary stream (actually, a property), and since primary stream sent no events before, I check if it emitted empty array. I also wanted to clarify what does the first line in the test case and a difference with the last line. As I see it, first one checks `Kefir.bufferWhileBy` and last one checks `stream.bufferWhileBy(otherStream)`. Is what I see correct?)

test/specs/buffer-while-by.coffee
```diff
@@ -20,12 +20,24 @@ describe 'bufferWhileBy', ->
...
       b = stream()
       expect(a.bufferWhileBy(b)).toEmit ['<end>'], -> send(a, ['<end>'])
 
+    it 'should flush empty buffer and then end when primary ends (w/ {emitEmpty: true})', ->
+      expect(send(stream(), ['<end>']).bufferWhileBy(stream(), {emitEmpty: true})).toEmit [{current: []}, '<end:current>']
+      a = stream()
+      b = stream()
+      expect(a.bufferWhileBy(b, {emitEmpty: true})).toEmit [[], '<end>'], -> send(a, ['<end>'])
+
+    it 'should flush empty buffer and then end when secondary ends (w/ {flushOnChange: true, emitEmpty: true})', ->
```
rpominov: The test seems ok, but the title should be "should flush empty buffer when secondary emits false ...".

rpominov:
> Also I wanted to clarify what does the first line in the test case and a difference with the last line. As I see it, first one checks Kefir.bufferWhileBy and last one checks stream.bufferWhileBy(otherStream). Is what I see correct?)

Not sure what you mean, we don't have `Kefir.bufferWhileBy`

It's all now looks good to me except for the test title.

shamansir:
> Not sure what you mean, we don't have Kefir.bufferWhileBy

True, I somehow thought every stream instance-bound method has a "backup" as a static method. Though, yet I found `exports.stream`/`prop` in `test-helpers`, I feel there is a major difference between the first line and the last line in the test case, but I'd like to clarify, what is it? Or does first line tests if current value was properly emitted, and last one does not? I mean these two:

```coffeescript
expect(stream().bufferWhileBy(send(prop(), [true, false]), {flushOnChange: true, emitEmpty: true})).toEmit [{current: []}]
expect(a.bufferWhileBy(b, {flushOnChange: true, emitEmpty: true})).toEmit [[]], -> send(b, [true, false])
```

rpominov:
> Or does first line tests if current value was properly emitted

Yeah, basically

In `expect(observable).toEmit(expectedLog, callback)` the callback is optional. It subscribes to `observable` and save events to a log; then if callback provided, calls the callback; then unsubscribe end compare log with `expectedLog`. So without callback it only captures current values.

It should be `send(prop(), [false])` in first line, btw. Both `send(prop(), [true, false])` and `send(prop(), [false])` create a property with current value `false`.

shamansir:
:pray:

rpominov:
Merged, thanks! 

I'll release a new version soon.

rpominov:
Just published `2.8.0`. Sorry it took so long, busy week :bee:

shamansir:
Thank you a lot! Will it soon be at `npm` or is it there already? No rush, just asking :)

rpominov:
Yeah, it's on npm :)

shamansir:
Great! Thanks again! :)

PR #135, opened 2015-08-27
Fix #134

### LINKED ISSUE #134: TypeError: Cannot read property 'shift' of null
In our application, we catch all uncaught exceptions and have the client log them to our server automatically. We've been seeing this many times a day from many users without much else in common:

Error logged: TypeError: Cannot read property 'shift' of null 
    at mixin.\_init.\_$shiftBuff ([src/one-source/delay.js:8](https://github.com/rpominov/kefir/blob/0339735b451a0ed0f93f59e0563f04dc4f8a390c/src/one-source/delay.js#L8))

I've seen it happen a few times randomly, but I haven't been able to dependably reproduce the issue until today. My reproduction steps involved triggering a few certain bugs in the application while having Chrome's debugger set to pause on exceptions, and then this exception happens afterward. (I don't assume this is how users are triggering it.) To get more information, I modified delay.js's code to have a few lines added:

```javascript
const mixin = {
  //...
  _free() {
    this.__free_stack = new Error().stack; // +
    this._buff = null;
    this._$shiftBuff = null;
  },
  //...
  _handleEnd() {
    if (this._activating) {
      this.__immediate_end = true; // +
      this._emitEnd();
    } else {
      setTimeout(() => {
        this.__timeout_end = true; // +
        this._emitEnd();
      }, this._wait);
    }
  }

};
```

At the moment of the exception, `this.__timeout_end` was true, and `this.__free_stack` contained:

Error 
    at AnonymousObservable._free ([src/one-source/delay.js:12](https://github.com/rpominov/kefir/blob/0339735b451a0ed0f93f59e0563f04dc4f8a390c/src/one-source/delay.js#L12)) 
    at AnonymousObservable._clear ([src/patterns/one-source.js:52](https://github.com/rpominov/kefir/blob/0339735b451a0ed0f93f59e0563f04dc4f8a390c/src/patterns/one-source.js#L52)) 
    at AnonymousObservable._emitEnd ([src/property.js:41](https://github.com/rpominov/kefir/blob/0339735b451a0ed0f93f59e0563f04dc4f8a390c/src/property.js#L41)) 
    at [src/one-source/delay.js:29](https://github.com/rpominov/kefir/blob/0339735b451a0ed0f93f59e0563f04dc4f8a390c/src/one-source/delay.js#L29)

It looks like the function assigned to `this._$shiftBuff` is being called after `this._emitEnd`. It looks like the code is relying on setTimeout callbacks with the same wait value being called in the order they're set, but this is not always the case: http://stackoverflow.com/a/5998338/1289657. If you view the example jsfiddle in Chrome, then you can easily trigger an error by hitting the pause and play buttons on the debugger a few times if one doesn't happen on its own.

Here's an example test case which replaces setTimeout with a function more prone to executing events out of order and fails about 20% of the time with the "Cannot read property 'shift' of null" error.

```javascript
var Kefir = require('kefir');

(function() {
  var originalST = global.setTimeout;
  global.setTimeout = function() {
    if (Math.random() > 0.5) {
      return originalST.apply(this, arguments);
    } else {
      var _this = this, _arguments = arguments;
      originalST(function() {
        return originalST.apply(_this, _arguments);
      }, 1);
    }
  };
})();

Kefir.sequentially(0, [1,2,3]).delay(0).log();
```

AgentME:
This fixes #134 and works around browser bugs that cause setTimeout callbacks to not be executed in the correct order.

Currently in `delay`, there is a buffer that holds values to emit. When the source stream emits a value, the value gets put into the buffer, and setTimeout is used to schedule pulling one item out of the buffer. When the source stream ends, setTimeout is used to schedule ending the delayed stream. The issue is that if the browser executes the setTimeout callback for ending the stream before executing the remaining callbacks for flushing the buffer, then the stream ends before the buffer has been cleared, and then an exception is thrown when we try to flush the buffer after the stream has ended.

This change makes it so when the source stream ends, we place a special sentinel value in the buffer, and only end the stream when we flush this value out. Because the end event is now in the value buffer, we don't have to rely on the browser to stop it from being processed out of turn. This unifies the handling of how value and end events are delayed, and happens to work around browser setTimeout bugs.

### DIFF
--- a/test/test-helpers.coffee
+++ b/test/test-helpers.coffee
@@ -145,18 +145,18 @@ beforeEach ->
 
     # Test with setTimeout not guaranteeing order. See
     # https://github.com/rpominov/kefir/issues/134
-    toEmitOverShakyTime: (expectedLog, cb, timeLimit = 1000) ->
+    toEmitOverShakyTime: (stOrder, expectedLog, cb, timeLimit = 1000) ->
       log = null
       exports.withFakeTime (tick) =>
         originalST = global.setTimeout
         global.setTimeout = ->
-          if Math.random() > 0.5
-            originalST.apply this, arguments
-          else
+          if stOrder.shift()
             _arguments = arguments
             originalST =>
               originalST.apply this, _arguments
             , 0
+          else
+            originalST.apply this, arguments
         log = exports.watchWithTime(@actual)
         cb?(tick)
         tick(timeLimit)
--- a/test/test-helpers.coffee
+++ b/test/test-helpers.coffee
@@ -1,5 +1,6 @@
 Kefir = require("../dist/kefir")
 sinon = require('sinon')
+Combinatorics = require('js-combinatorics')
 
 Kefir.DEPRECATION_WARNINGS = false;
 
@@ -67,6 +68,36 @@ exports.prop = ->
 
 exports.stream = ->
   new Kefir.Stream()
+
+# This will run the callback timeoutCount^timeoutCount times. Each time, the
+# callback will be expected to call setTimeout exactly timeoutCount times. The
+# setTimeout callbacks may be scheduled as normal, or with an up to
+# timeoutCount-1 extra delay added. This will test all combinations. This is to
+# test compatibility with browser bugs that reorder setTimeout callbacks. See
+# https://github.com/rpominov/kefir/issues/134 for more information.
+exports.shakyTimeTest = (timeoutCount, cb) ->
+  delayValues = [0..timeoutCount-1]
+  Combinatorics.baseN(delayValues, timeoutCount).forEach (order) =>
+    expectToEmitOverShakyTime = (stream, expectedLog, cb, timeLimit = 10000) =>
+      log = null
+      exports.withFakeTime (tick) =>
+        originalST = global.setTimeout
+        global.setTimeout = sinon.spy ->
+          delay = order.shift()
+          if delay == 0
+            originalST.apply this, arguments
+          else
+            _arguments = arguments
+            originalST =>
+              originalST.apply this, _arguments
+            , delay
+        log = exports.watchWithTime(stream)
+        cb?(tick)
+        tick(timeLimit)
+        expect(global.setTimeout.callCount).toBe(timeoutCount)
+        global.setTimeout = originalST
+      expect(log.map ([time, value]) -> value).toEqual(expectedLog)
+    cb expectToEmitOverShakyTime
 
 exports.withFakeTime = (cb) ->
   clock = sinon.useFakeTimers(10000)
@@ -142,27 +173,6 @@ beforeEach ->
         tick(timeLimit)
       @message = -> "Expected to emit #{jasmine.pp(expectedLog)}, actually emitted #{jasmine.pp(log)}"
       @env.equals_(expectedLog, log)
-
-    # Test with setTimeout not guaranteeing order. See
-    # https://github.com/rpominov/kefir/issues/134
-    toEmitOverShakyTime: (stOrder, expectedLog, cb, timeLimit = 1000) ->
-      log = null
-      exports.withFakeTime (tick) =>
-        originalST = global.setTimeout
-        global.setTimeout = ->
-          if stOrder.shift()
-            _arguments = arguments
-            originalST =>
-              originalST.apply this, _arguments
-            , 0
-          else
-            originalST.apply this, arguments
-        log = exports.watchWithTime(@actual)
-        cb?(tick)
-        tick(timeLimit)
-        global.setTimeout = originalST
-      @message = -> "Expected to emit #{jasmine.pp(expectedLog)}, actually emitted #{jasmine.pp(log)}"
-      @env.equals_(expectedLog, log.map ([time, value]) -> value)
 
     toActivate: (obss...) ->
 
--- a/test/test-helpers.coffee
+++ b/test/test-helpers.coffee
@@ -143,6 +143,27 @@ beforeEach ->
       @message = -> "Expected to emit #{jasmine.pp(expectedLog)}, actually emitted #{jasmine.pp(log)}"
       @env.equals_(expectedLog, log)
 
+    # Test with setTimeout not guaranteeing order. See
+    # https://github.com/rpominov/kefir/issues/134
+    toEmitOverShakyTime: (expectedLog, cb, timeLimit = 1000) ->
+      log = null
+      exports.withFakeTime (tick) =>
+        originalST = global.setTimeout
+        global.setTimeout = ->
+          if Math.random() > 0.5
+            originalST.apply this, arguments
+          else
+            _arguments = arguments
+            originalST =>
+              originalST.apply this, _arguments
+            , 0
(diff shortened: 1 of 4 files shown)


test/test-helpers.coffee
```diff
@@ -143,6 +143,27 @@ beforeEach ->
...
       @env.equals_(expectedLog, log)
 
+    # Test with setTimeout not guaranteeing order. See
+    # https://github.com/rpominov/kefir/issues/134
+    toEmitOverShakyTime: (expectedLog, cb, timeLimit = 1000) ->
+      log = null
+      exports.withFakeTime (tick) =>
+        originalST = global.setTimeout
+        global.setTimeout = ->
+          if Math.random() > 0.5
```
rpominov: Not sure if randomness in tests is a good idea. What if we pass an option to `toEmitOverShakyTime` to tell what path to take?

```coffeescript
for reverse in [true, false]
  a = stream()
  expect(a.delay(0)).toEmitOverShakyTime reverse, [i, '<end>'], (tick) ->
    send(a, [i])
    send(a, ['<end>'])
```

```diff
- if Math.random() > 0.5
+ if reverse
```

AgentME: Each run calls setTimeout twice, and I didn't want to couple `toEmitOverShakyTime` with that. I also wanted it to test the regular cases where both setTimeout calls are not or are delayed together, which shouldn't ever fail but act as a good control for the test really just testing `toEmitOverShakyTime` itself.

I've put in another commit that removes the non-determinism and replaces it with code closer to your suggestion.

rpominov:
Ok. I didn't quite understood how `toEmitOverShakyTime` works at first, and thought that making it deterministic will be easier...

Is it important for you to get the fix soon? In this case I'd fall back to `Math.random` pattern, and merge. We can figure out better solution for tests later.

AgentME:
The strategy I used would get stupidly verbose if `toEmitOverShakyTime` was ever used with more than two setTimeouts, and requires us to be diligent about listing out all of the test cases. If we add [js-combinatorics](https://www.npmjs.com/package/js-combinatorics) as a dev-dependency, we could change

```diff
-      for order in [[false, false], [false, true], [true, false], [true, true]]
+      for order in Combinatorics.baseN([false, true], 2).toArray()
```

Obviously only a small improvement in this case, but it makes it more obvious we didn't miss a case, and is scalable to possible larger tests in the future.

Ideally it would be good to move that line into a test helper function, but it couldn't be moved into `toEmitOverShakyTime` because a new stream has to be created each time and then have `.delay(0)` called on it. We could make a test helper function that takes a callback and a timeout count (2 here), it automatically calls the callback the correct number of times with the modified setTimeout function, and the callback calls `expect(a.delay(0)).toEmitInTime` itself, but that breaks down because `toEmitInTime` uses the sinon fake timer, which overrides our setTimeout override. We need to place our setTimeout override only after sinon has placed its fake timer. The only way I can think of to allow the concept of the timing orders to be completely encapsulated in the test helper makes the test look like this:

```coffeescript
shakyTimeTest 2, [4, '<end>'],
  -> stream()
  (a) -> a.delay(0)      
  (a, tick) ->
    send(a, [4])
    send(a, ['<end>'])
```

which is pretty ugly and really obfuscates what the test is doing. I think the more explicit test which handles calling `Combinatorics.baseN` itself and passes an order to `toEmitOverShakyTime` is much better.

I'm fine if this waits a few days.

rpominov:
Hm, we also can make helper API to be something like this:

```coffeescript
shakyTimeTest 2, (expectToEmitOverShakyTime) ->
  a = stream()
  expectToEmitOverShakyTime a.delay(0), [i, '<end>'], (tick) ->
    send(a, [i])
    send(a, ['<end>'])
```

Here `shakyTimeTest` expects `setTimeout` to be called 2 times, and will call its callback `2*2=4` times, providing each time a new `expectToEmitOverShakyTime` set up with a different order. Still don't like that we have to specify how many times `setTimeout` will be called though.

Another direction that can be fruitful is to see if we could interop somehow with the mock. E.g. perhaps it's possible to change/replace `clock.tick` implementation — make it call handlers, that set up for the same time, in the reverse order. Then we could make the API with the `reverse` option as I suggested at first.

AgentME:
>`2*2=4` times

Clarification: I'm intending `2^2=4` times. If three setTimeouts were used, then it would be called `2^3=8` times with the orders FFF, FFT, FTF, FTT, ... . Oh, actually it would make more sense to do `3^3=9` with 0,1,2 instead of false,true to get fuller coverage in that case. But I don't think it'd be necessary for us to do with 3.

>Still don't like that we have to specify how many times `setTimeout` will be called though.

We could make it count how many times setTimeout is called on the first run, and then generate orderings from that to use for the next runs (and error if any runs call it a different number of time). Could be overkill though, and could lead to the test callback running exponentially more times than intended if setTimeout was called too many times. You could consider it part of the test that it checks that setTimeout is only called the given number of times within the callback. (Not that it's enforced at the moment.)

AgentME:
I've implemented your `shakyTimeTest` suggestion.

rpominov:
> Clarification: I'm intending 2^2=4 times.

Yeah, I meant both `2` is a parameter in `2*2` — `x*x=result` :)

> Oh, actually it would make more sense to do 3^3=9

Don't quite get it yet, but maybe you're right.

-----

Although I don't sure we have to test all combination. I think only two runs is enough: a normal one & one with "simultaneous"  setTimeout's handlers called in reverse order. In this diagram letters are handlers:

```
            a  bc  def
run 1: -----•--•---•---

            a  cb  fed
run 2: -----•--•---•---
```

But in order to do that we need a different approach: replace `clock.tick()` in the mock / use a different mock that supports that feature / check if our mock supports it, etc. I could investigate this area at weekend. 

I think https://github.com/sinonjs/lolex is the mock we use now (we use sinon, which uses this).

AgentME:
Hm, `shakyTimeTest` does have some redundancy right now. The case of both timeouts being delayed by 1 is redundant with the case of both timeouts not being delayed, and the case of both timeouts executing with no delay is redundant with the case of the second timeout being delayed by 1. For code where all setTimeouts are scheduled at the same time for the same delay, we don't ever need to test delaying multiple timeouts by the same value. Using `Combinatorics.permutation` instead of `Combinatorics.baseN` would fix that, and make it so for two timeouts, only two runs are done which is all that's necessary.

I think adding arbitrary delays to timeout calls meets our needs here, though having lolex support a reversed tick function would be a neat feature for them to have, and let us do quick non-exhaustive tests with more timeouts.

AgentME:
Oh I see the disconnect between us. You're thinking of cases where setTimeouts are scheduled at multiple times, and I've been only thinking about exhaustive tests of code where all setTimeouts are scheduled for the same time. I agree with you now, getting lolex to have a flag for doing concurrent timeouts in reverse order would be more general and let us test more. It wouldn't be as exhaustive as my code, but my code only works properly for a specific case of all timeouts being set together, and I don't think that specialization is needed.

rpominov:
I've manually merged the fix. Thank you for the PR!

PR #139, opened 2015-09-04
Property: set currentEvent before dispatching it

olivierguerriat:
see https://github.com/rpominov/kefir/issues/127

### DIFF
--- a/src/property.js
+++ b/src/property.js
@@ -16,19 +16,19 @@ inherit(Property, Observable, {
 
   _emitValue(value) {
     if (this._alive) {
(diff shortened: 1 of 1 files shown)


rpominov:
Could you also write a test case for this?

rpominov:
Thanks!

olivierguerriat:
Thank you for writing the tests, it would have taken me much longer as I am not familiar with Kefir's test helpers nor with CoffeeScript.

rpominov:
I see, np :)

PR #154, opened 2015-09-25
Compatibility with ES7 observables

lautis:
Use [ES7 observables](https://github.com/zenparsing/es-observable) as Kefir streams and allow Kefir streams to be converted to ES7 observables.

To initialise Kefir streams from ES7 observables, there is a new method `Kefir.from`. This could also be `Kefir.fromObservable`, but I've tried to conform with the ES7 observable spec.

To make Kefir streams compatible with observables, streams now have a new method for subscription: `stream.subscribe`. This takes an ES7-like observer as an argument. Unlike other methods, subscriptions initialised via subscribe will always end after first error. This is mandated by ES7 observable spec.

There is a similar PR for Bacon.js, and the upcoming version of RxJS will be also compatible with ES7 observables. Use [ES7 observables](https://github.com/zenparsing/es-observable) as Kefir streams and allow Kefir streams to be converted to ES7 observables.

To initialise Kefir streams from ES7 observables, there is a new method `Kefir.from`. This could also be `Kefir.fromObservable`, but I've tried to conform with the ES7 observable spec.

To make Kefir streams compatible with observables, streams now have a new method for subscription: `stream.subscribe`. This takes an ES7-like observer as an argument. Unlike other methods, subscriptions initialised via subscribe will always end after first error. This is mandated by ES7 observable spec.

There is a similar PR for Bacon.js, and the upcoming version of RxJS will be also compatible with ES7 observables. As a bonus, these would be interoperable with Kefir.

### DIFF
--- a/src/observable.js
+++ b/src/observable.js
@@ -101,6 +101,22 @@ extend(Observable.prototype, {
     return this._on(ANY, fn);
   },
 
+  subscribe(observer) {
+    let stream = this.takeErrors(1);
+    let fn = function(event) {
+      if (event.type === "value" && observer.next) {
+        observer.next(event.value);
+      } else if (event.type == "error" && observer.error) {
+        observer.error(event.value);
+      } else if (event.type === "end" && observer.complete) {
+        observer.complete(event.value);
+      }
+    }
+
+    stream.onAny(fn);
+    return () => stream.offAny(fn);
+  },
+
   offValue(fn) {
     return this._off(VALUE, fn);
   },
@@ -166,6 +182,7 @@ extend(Observable.prototype, {
 
 });
 
+Observable.prototype[require('./utils/symbol')('observable')] = function() { return this; }
 
 // extend() can't handle `toString` in IE8
 Observable.prototype.toString = function() {
--- a/src/observable.js
+++ b/src/observable.js
@@ -163,13 +163,14 @@ extend(Observable.prototype, {
     }
 
     return this;
-  }
+  },
 
+  toESObservable() {
+    return new ESObservable(this);
+  }
 });
 
-Observable.prototype[require('./utils/symbol')('observable')] = function() {
-  return new ESObservable(this);
-}
+Observable.prototype[require('./utils/symbol')('observable')] = Observable.prototype.toESObservable;
 
 // extend() can't handle `toString` in IE8
 Observable.prototype.toString = function() {
--- a/src/observable.js
+++ b/src/observable.js
@@ -2,6 +2,7 @@ const {extend} = require('./utils/objects');
 const {VALUE, ERROR, ANY, END} = require('./constants');
 const {Dispatcher, callSubscriber} = require('./dispatcher');
 const {findByPred} = require('./utils/collections');
+const ESObservable = require('./es-observable');
 
 
 
@@ -101,22 +102,6 @@ extend(Observable.prototype, {
     return this._on(ANY, fn);
   },
 
-  subscribe(observer) {
-    let stream = this.takeErrors(1);
-    let fn = function(event) {
-      if (event.type === "value" && observer.next) {
-        observer.next(event.value);
-      } else if (event.type == "error" && observer.error) {
-        observer.error(event.value);
-      } else if (event.type === "end" && observer.complete) {
-        observer.complete(event.value);
-      }
-    }
-
-    stream.onAny(fn);
-    return () => stream.offAny(fn);
-  },
-
   offValue(fn) {
     return this._off(VALUE, fn);
   },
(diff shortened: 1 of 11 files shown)


rpominov:
Hey @lautis! Thanks a lot for the PR! I didn't followed the ES proposal discussions for some time and lost track a bit. I think there were some other inconsistencies beside ending on error. I will need to look closely at the proposal again.

I'll return to this soon. Also very curious what others think about it!

rpominov:
I think, at this point it makes sense to only make Kefir observables be convertible to/from ES observables, but not to make Kefir implement ES observable spec. 

I mean we should not try to add methods from ES observable spec like `Kefir.Observable.prototype.from()` and `Kefir.Observable.prototype.subscribe()`. We should either add them all (the Observable constructor, `forEach`, `of` + any future methods), and maintain compatibility with the spec, or don't try to do this at all. The former looks like a lot of work and discussions.

So I think we should add only two following methods:
  - something like `Kefir.Observable.prototype.fromESObservable()`
  - and `Kefir.Observable.prototype[Symbol.observable]()` that returns an object with `subscribe` method (what @zenparsing suggests in https://github.com/baconjs/bacon.js/issues/633#issuecomment-142414957)

Not sure about `fromESObservable` name, although I don't like `fromObservable` because it can be confusing as we have something called `Observable` in Kefir already.

Also maybe it makes sense to add `obs.toESObservable` as an alias for `obs[Symbol.observable]`.

src/utils/symbol.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function(key) {
+  if (Symbol && Symbol[key]) {
```
rpominov: We should probably do `typeof Symbol !== 'undefined'` here, otherwise this will trow in case `Symbol` not defined as a global.

test/specs/from.coffee
```diff
@@ -0,0 +1,30 @@
+Observable = require('zen-observable')
+Rx = require('@reactivex/rxjs')
```
rpominov: Hm, how does this `require` work?

lautis: It's a [scoped package](https://docs.npmjs.com/misc/scope) – a relatively new feature in NPM that seems to be only available for 0.12 and newer nodes.

rpominov: I see, we should add it to `package.json` then. I thought at first that `@` makes it somehow work without installing :)

lautis: Yep, added.

lautis:
Extracted interoperability to separate class and changed the name of static method to consume ES7 observables to `fromESObservable`.

Re: `Kefir.from`, that might make more sense if it could be a general method to turn objects into a stream. But as Kefir doesn't (won't?) have an array/iterable to a stream method, there aren't really other use cases that could be easily handled. `Kefir.fromESObservable` seems to be the least confusing name.

src/observable.js
```diff
@@ -162,10 +163,14 @@ extend(Observable.prototype, {
     }
 
     return this;
-  }
+  },
 
+  toESObservable() {
```
rpominov: I think we should add this two methods to the prototype in `index.js` file, for consistency with how other methods are added.

rpominov:
Not sure yet what we should do with exceptions from observer methods. 

And not 100% sure that it's ok to return not a zalgo-safe observable. Although you're probably right that spec doesn't require that.

I wonder what @blesh and @zenparsing think about it. I'd wait for their answers in the Bacon issue.

lautis:
I looked at the exception handling a bit more, especially on what happens in Kefir after an exception is thrown. As Kefir doesn't catch any exceptions (which does help a lot for performance), streams might end up in inconsistent state as the depending on whether the ES stream implementation unsubscribes properly or not.

The code for exception handling is in https://github.com/lautis/kefir/tree/es-observable-exceptions.

I'd definitely like to hear what the other implementations expect to happen on exception. It could also mean that `Kefir.fromESObservable` should unsubscribe on error.

blesh:
In RxJS, all errors are caught and sent down the error handler path. If there is no error handler, it throws. Related: any error in a handler will also throw.

rpominov:
@blesh Yeah, the problem this is not the case in Kefir (and Bacon). In Kefir there is no try..catch blocks in the code. We intentionally not catching exceptions and just allow app to crash hard.

The question is should we still catch exceptions in the ES Observable adapter? Does spec require that? I mean exceptions thrown from `observer.next()`, `observer.error()`, and `observer.complete()`.

blesh:
They shouldn't throw. You should be okay.

rpominov:
Ok, I think it makes sense to not catch errors at this point. We'll be able to change this behavior in the future after people start use it, and we'll get some feedback.

Gona release a new version with this a bit later.

Thanks a lot for the PR, @lautis!

PR #164, opened 2015-10-19
Fixed misspelling of Promise as 'Propmise'.

sartaj:
https://github.com/rpominov/kefir/issues/163


rpominov:
Thanks for the find!

PR #169, opened 2015-11-23
use constants for consistency

aj0strow:
As per https://github.com/rpominov/kefir/issues/168

I was reading the source to learn. Thanks for the wonderful package!


rpominov:
Thanks :+1:

Issue #161, opened 2015-10-11
takeNth?


elsehow:
Hi!

I've written a small wrapper to [take every nth value](https://github.com/elsehow/kefir-takenth) of a Kefir stream. I have no idea if my solution is eloquent, but would there be any interest in integrating this feature into Kefir?

rpominov:
Hey!

Not sure, I've personally never needed this. I'd like to see what others think.
What is your use case, btw? What do you need it for?

Btw, I'd implement it a bit easier:

```js
function takeNth (stream, n) {
  let count = 0
  return stream.filter(() => {
    count++
    return count % n == 0
  })
}
```

This method is inevitably stateful so it's ok to store some state in a closure variable.

elsehow:
Nice solution! I've updated my package.

My use case is taking buffers of EEG data flowing over a serial port:

```javascript
serialData.slidingWindow(512,512).takeNth(512).map(fft)
```

rpominov:
I see. `.slidingWindow(512,512).takeNth(512)` probably could be replaced with `.bufferWithCount(512)` [if we'd have `bufferWithCount`](https://github.com/rpominov/kefir/issues/132).

elsehow:
Agreed. I'll try to work on those and maybe issue a PR. Thanks

rpominov:
:+1:

PR #170, opened 2015-11-27
added bufferByCount method to Observable

elsehow:
added tests in test/specs

updated docs in docs-src

adding this method was mentioned in #161 

`bufferByTime` could also be nice, though i haven't implemented it in this PR

### DIFF
--- a/docs-src/descriptions/one-source.jade
+++ b/docs-src/descriptions/one-source.jade
@@ -654,7 +654,6 @@ div
 
 
 
-
 +descr-method('sliding-window', 'slidingWindow', 'obs.slidingWindow(max, [min])').
   Will emit arrays containing the last #[b n] values from the #[b obs] observable,
   where #[b n] is between #[b max] and #[b min] arguments.
@@ -679,6 +678,7 @@ pre(title='events in time').
   result:  -----------------•--------•--------•--------•X
                         [1,2]  [1,2,3]  [2,3,4]  [3,4,5]
 div
+
 
 
 
@@ -709,6 +709,35 @@ pre(title='events in time').
   source:  ---1---2---3---4---5 X
   result:  -----------•--------•X
                 [1,2,3]    [4,5]
+div
+
+
+
++descr-method('buffer-with-count', 'bufferWithCount', 'obs.bufferWithCount(count)').
+  Buffers all values from #[b obs] observable, and flushes the buffer
+  every time #[b n] values have been passed through. Also flushes the buffer before end, if the buffer is not empty.
+
+p.
+  If #[tt options.flushOnEnd] is #[tt false], the buffer
+  won't be flushed when the main observable ends.
+
+pre.javascript(title='example')
+  :escapehtml
+    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5]);
+    var result = source.bufferWithCount(2)
+    result.log();
+
+pre(title='console output')
+  :escapehtml
+    > [sequentially.bufferWithCount] <value> [1, 2]
+    > [sequentially.bufferWithCount] <value> [3, 4]
+    > [sequentially.bufferWithCount] <value> [5]
+    > [sequentially.bufferWithCount] <end>
+
+pre(title='events in time').
+  source:  --------1--------2--------3--------4--------5 X
+  result:  -----------------•-----------------•---------•X
+                        [1,2]             [3,4]       [5]X
 div
 
 
--- a/src/index.js
+++ b/src/index.js
@@ -264,6 +264,13 @@ Observable.prototype.bufferWhile = function(fn, options) {
   return bufferWhile(this, fn, options);
 };
 
+// (Stream, number) -> Stream
+// (Property, number) -> Property
+const bufferWithCount = require('./one-source/buffer-with-count');
+Observable.prototype.bufferWithCount = function(count, options) {
+  return bufferWithCount(this, count, options);
+};
+
 // (Stream, Function) -> Stream
 // (Property, Function) -> Property
 const transduce = require('./one-source/transduce');
--- a/test/specs/buffer-with-count.coffee
+++ b/test/specs/buffer-with-count.coffee
@@ -1,0 +1,110 @@
+{stream, prop, send, Kefir} = require('../test-helpers.coffee')
+
+describe 'bufferWithCount', ->
+
+  describe 'stream', ->
+
+    it 'should return stream', ->
+      expect(stream().bufferWithCount(1)).toBeStream()
+
+    it 'should activate/deactivate source', ->
+      a = stream()
+      expect(a.bufferWithCount(1)).toActivate(a)
+
+    it 'should be ended if source was ended', ->
+      expect(send(stream(), ['<end>']).bufferWithCount(1)).toEmit([
+        '<end:current>'
+      ])
+
+    it '.bufferWithCount(1) should work correctly', ->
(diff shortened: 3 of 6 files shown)


rpominov:
Cool, thanks. A few notes:

Lets rename it to `bufferWithCount` similar to how it named in Rx and Bacon.

Lets make it emit incomplete buffer at end (even if the buffer is empty) similar to how other buffering methods in Kefir do.

rpominov:
Please remove empty file with name `2`.

docs-src/descriptions/one-source.jade
```diff
@@ -712,6 +712,30 @@ div
...
+pre(title='console output')
+  :escapehtml
+    > [sequentially.bufferByCount] <value> [1, 2]
+    > [sequentially.bufferByCount] <value> [3, 4]
+    > [sequentially.bufferByCount] <end>
+
+pre(title='events in time').
+  source:  --------1--------2--------3--------4--------5X
+  result:  -----------------•--------•--------•--------•X
+                        [1,2]             [3,4]         X
```
rpominov: It should be:
```
result:  -----------------•-----------------•---------X
                         [1,2]             [3,4]         X
```
(extra dots removed)

rpominov:
And we also probably need optional `options` argument with one option `flushOnEnd` similar to other buffering methods.

elsehow:
thanks for these notes. i have it emitting an incomplete buffer at end, but, unless i'm misunderstanding something, `bufferBy` doesn't emit on end if the buffer is empty. so, i made `bufferWithCount` act the same way.

rpominov:
> bufferBy doesn't emit on end if the buffer is empty.

I think it does, there is no checks for `buffer.length` in _flush or _handlePrimaryEnd in it https://github.com/rpominov/kefir/blob/master/src/two-sources/buffer-by.js#L15-L27

Also we need to add info about `options` to the doc, and update events diagram (see how it done in bufferBy)

image

Note the space before X in source stream, it's trying to illustrate that the flushing event inserted between 8 and X events of original stream, and Xs "happen at the same time".

```
source:  --------1--------2--------3--------4--------5 X
result:  -----------------•-----------------•---------•X
                      [1,2]             [3,4]       [5]X
```

test/specs/buffer-with-count.coffee
```diff
@@ -0,0 +1,107 @@
...
+        [1, 2, 3]
+        [4, 5]
+        '<end>'
+      ], -> send(a, [1, 2, 3, 4, 5,'<end>'])
+
+    it 'should not flush buffer on end if {flushOnEnd: false}', ->
+      a = stream()
+      expect(a.bufferWithCount(3, {flushOnEnd: false})).toEmit [
+        [1, 2, 3]
+        [4, 5]
```
rpominov: Weird, it shouldn't emit `[4, 5]` with `{flushOnEnd: false}`, right?

docs-src/descriptions/one-source.jade
```diff
@@ -713,6 +713,38 @@ div
...
+pre.javascript(title='example')
+  :escapehtml
+    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5]);
+    var result = source.bufferWithCount(2)
+    result.log();
+
+pre(title='console output')
+  :escapehtml
+    > [sequentially.bufferWithCount] <value> [1, 2]
+    > [sequentially.bufferWithCount] <value> [3, 4]
```
rpominov: `[5]` should be here

rpominov:
We also need link to the new method in sidebar in docs.

test/specs/buffer-with-count.coffee
```diff
@@ -0,0 +1,107 @@
...
+      expect(stream().bufferWithCount(1)).toBeStream()
+
+    it 'should activate/deactivate source', ->
+      a = stream()
+      expect(a.bufferWithCount(1)).toActivate(a)
+
+    it 'should be ended if source was ended', ->
+      expect(send(stream(), ['<end>']).bufferWithCount(1)).toEmit 
+      []
+      ['<end:current>']
```
rpominov: CoffeeScript is crazy. I don't understand what happening here. Where the second array go?

test/specs/buffer-with-count.coffee
```diff
@@ -0,0 +1,107 @@
...
+        [4, 5]
+        '<end>'
+      ], -> send(a, [1, 2, 3, 4, 5,'<end>'])
+
+    it 'errors should flow', ->
+      a = stream()
+      expect(a.bufferWithCount(3)).errorsToFlow(a)
+
+
+  describe 'property', ->
```
rpominov: Please also add flushOnEnd tests with property.

src/index.js
```diff
@@ -264,6 +264,13 @@ Observable.prototype.bufferWhile = function(fn, options) {
   return bufferWhile(this, fn, options);
 };
 
+// (Stream, number) -> Stream
+// (Property, number) -> Property
+const bufferWithCount = require('./one-source/buffer-with-count');
+Observable.prototype.bufferWithCount = function(count) {
```
rpominov: You don't pass `options` here, that is why `{flushOnEnd: false}` doesn't work probably.

test/specs/buffer-with-count.coffee
elsehow: sorry, no language should have let me do that :)

what I meant to do (and what i would expect to happen) is this:

```coffee
 it 'should be ended if source was ended', ->
    expect(send(stream(), ['<end>']).bufferWithCount(1)).toEmit([
        [],
        '<end:current>'
      ])
```

what i see now, and what passes the tests (now that i am passing in the options correctly), is this:

```coffee
    it 'should be ended if source was ended', ->
      expect(send(stream(), ['<end>']).bufferWithCount(1)).toEmit([
        { current: [] },
        '<end:current>'
      ])
```

i am confused about the `{ current }` value. my understanding was that only properties have the notion of current values, but here we're passing in a stream?

rpominov: > i am confused about the { current } value. my understanding was that only properties have the notion of current values, but here we're passing in a stream?

It depends on what we consider "current" value. In tests `{ current: something }` or `'<end:current>'` only means that the callback was called immediately before subscribing method returns, i.e.:

```js
obs.onValue(() => { log(1) });
log(2);

// logs
// 1
// 2
```
 
See also http://rpominov.github.io/kefir/#current-in-streams

rpominov: > no language should have let me do that :)

btw, I'm thinking of moving away from CoffeeScript, but it's just a lot of code to migrate...

elsehow: > In tests { current: something } or '<end:current>' only means that the callback was called immediately before subscribing method returns [..] See also http...

got it, this makes sense. thanks for the reference

> btw, I'm thinking of moving away from CoffeeScript, but it's just a lot of code to migrate...

i agree that it would be good to move away from CoffeeScript. transpiling the existing tests to js and cleaning up the results could be the easiest way.

rpominov:
Looks good now, please squash commits and we can merge this.

rpominov:
Sorry, I thought more about it and changed my mind. let's actually add check for `buf.length !== 0` in flush (same as in bufferWhile). See also https://github.com/rpominov/kefir/pull/172#discussion_r46769826

rpominov:
Thanks, @elsehow!

Issue #132, opened 2015-08-18
.bufferWithTimeOrCount (& .bufferWithCount?)


rpominov:
[in RxJS](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/bufferwithtimeorcount.md)
[in Bacon](https://github.com/baconjs/bacon.js#stream-bufferwithtimeorcount)

We probably need this. 

We also lack `.bufferWithTime` and `.bufferWithCount`, but those are relatively easy to replicate using existing API:

```js
// .bufferWithTime
foo.bufferBy(Kefir.interval(1000))

// .bufferWithCount
let i = 0
foo.bufferWhile(() => i++ % n !== 0)
```

Although having to store `i` state somewhere for `.bufferWithCount` also not very cool. Maybe we need this one too.

PRs are welcome! :dancer:

vikikamath:
:+1:

mcmathja:
I'd be interested in taking this. To clarify, expected behavior is that timer is reset to 0 after a flush initiated by count, but otherwise should run continuously between activation and deactivation?

rpominov:
@mcmathja Yeah, I think timer should be reset to 0 on each flush (no matter what initiated it), and we should start timer on activation, and stop & reset it on deactivation.

mcmathja:
I've got something up and running, but I had a question about how Kefir handle's inheritance: this definitely seems like a *single-activation* pattern, but we need to override `_onActivation` and `_onDeactivation` to properly handle the interval. There's no straightforward way to call up to the pattern's implementation of these methods once they're overridden - I considered using *time-based* instead, but it can only return streams, and adding a super implementation to inherits seems to call for a big piece of refactoring. In *to-property* and most of the *two-sources* streams, the activation / deactivation methods defined in the patterns are simply copied over and reimplemented with local modifications - would that be the preferred approach to doing this?

rpominov:
We should probably do it same way as in _to-property_ here. I know it's not perfect solution, but easiest one and shouldn't make much harm. I would't use inheritance at all if it would't be good for performance :sweat_smile:

mcmathja:
Haha, all right! :stuck_out_tongue_winking_eye:

One other question, actually. I was reading through [Bacon's source](https://github.com/baconjs/bacon.js/blob/master/src/buffer.js) for this and they handle scheduling somewhat differently. Assuming the buffer is never flushed because the *count* limit is reached, the processing order ends up being:

[...]

PR #172, opened 2015-12-06
Adds bufferWithTimeOrCount

mcmathja:
Implements #132.

### DIFF
--- a/docs-src/descriptions/one-source.jade
+++ b/docs-src/descriptions/one-source.jade
@@ -726,28 +726,29 @@ p.
 
 pre.javascript(title='example')
   :escapehtml
-    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5, 6]);
-    var result = source.bufferWithTimeOrCount(310, 10);
+    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5, 6, 7, 8]);
+    var result = source.bufferWithTimeOrCount(330, 10);
     result.log();
 
 pre(title='console output')
   :escapehtml
     > [sequentially.bufferWithTimeOrCount] <value> [1, 2, 3]
     > [sequentially.bufferWithTimeOrCount] <value> [4, 5, 6]
+    > [sequentially.bufferWithTimeOrCount] <value> [7, 8]
     > [sequentially.bufferWithTimeOrCount] <end>
 
 pre(title='events in time').
-  source:  ---1---2---3---4---5---6X
-  result:  -----------•-----------•X
-                [1,2,3]     [4,5,6]
+  source:  ---1---2---3---4---5---6---7---8 X
+  result:  ------------•------------•------•X
+                [1,2,3]       [4,5,6]  [7,8]
 
 p.
   Limited by count:
 
 pre.javascript(title='example')
   :escapehtml
-    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5, 6]);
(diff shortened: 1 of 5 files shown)


docs-src/descriptions/one-source.jade
```diff
@@ -714,6 +714,58 @@ div
...
+pre(title='console output')
+  :escapehtml
+    > [sequentially.bufferWithTimeOrCount] <value> [1, 2, 3]
+    > [sequentially.bufferWithTimeOrCount] <value> [4, 5, 6]
+    > [sequentially.bufferWithTimeOrCount] <end>
+
+pre(title='events in time').
+  source:  ---1---2---3---4---5---6X
+  result:  -----------•-----------•X
+                [1,2,3]     [4,5,6]
```
rpominov: We should move `result`'s points a bit to show that flushes happen not at the same time with events in `source`, also the second point is the "flush on end" and should be shown a bit differentially. Let's modify it like this:

```js
var source = Kefir.sequentially(100, [1, 2, 3, 4, 5, 6, 7, 8]);
var result = source.bufferWithTimeOrCount(330, 10);
```
```
  source:  ---1---2---3---4---5---6---7---8 X
  result:  ------------•------------•------•X
                 [1,2,3]      [4,5,6]  [7,8]
```

src/one-source/buffer-with-time-or-count.js
```diff
@@ -0,0 +1,63 @@
...
+    this._buff = [];
+  },
+
+  _free() {
+    this._$onTick = null;
+    this._buff = null;
+  },
+
+  _flush() {
+    if (this._buff !== null && this._buff.length !== 0) {
```
rpominov: Let's not check for `length` here, and simply emit empty arrays when buffer is empty. This is necessary in some cases to know when buffer flushes, and it's easy to filter out those empty flushes with `.filter(x => x.length > 0)` (if the code that handles arrays sensitive to their length).

We already do this in all other buffering methods.

mcmathja: Sure - do you want to still avoid flushing an empty buffer on end, though? This isn't a common case with the other multi-source buffer methods because end typically hits right after some value. Here though, since we flush the buffer immediately any time count is reached, we'll emit an empty array on end whenever the source observable has `#events in last stretch % count = 0`. For instance, case 2 in the docs will become:

```
  source:  ---1---2---3---4---5---6---7---8  X
  result:  -------•-------•-------•-------•  X
              [1,2]   [3,4]   [5,6]   [7,8] []
```

This would also occur with the other buffer methods if end arrived substantially later than the final value, but it's a much more common corner case here since we handle the flush condition synchronously before processing end.

docs-src/descriptions/one-source.jade
mcmathja: :thumbsup:

src/one-source/buffer-with-time-or-count.js
rpominov: Hm, I think you're right. That makes sense to do check that `.lenght !== 0` in "on end" flush. I can't imagine a use case where it would be necessary to always have that "on end" flush. And we actually do that check in `.bufferWhile()` (we always check in there, but it can be empty only on end).

rpominov:
Could you update it to the current master?

rpominov:
All looks good now, thanks!
Will release a new version with this and `.bufferWithCont()` later this weak.

PR #176, opened 2016-01-01
Remove skip while

askmatey:
See 0d62caa99110c9569d4efa5f886c6cbcadd54c75

### DIFF
--- a/docs-src/descriptions/one-source.jade
+++ b/docs-src/descriptions/one-source.jade
@@ -268,9 +268,6 @@ pre(title='events in time').
   result: -------3---2X
 div
 
-p.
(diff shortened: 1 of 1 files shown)


rpominov:
Thanks, did't catch that. 

I'd wouldn't update `bacon-vs-kefir-api.md` though. It needs much more fixes than that. Currently it's a comparison of API's of two specific versions of libraries:

> Bacon v0.7.52 and Kefir v2.0.0

Please leave the change only in docs and squash commits, and we can merge this.

askmatey:
done deal

rpominov:
Ok, thanks.

Issue #178, opened 2016-01-13
fromNodeStream


aj0strow:
Like `fromNodeCallback` and `fromPromise`. 

It's nice to lift node streams into kefir streams. When a stream is in object mode, it's basically a kefir observable anyway with data, error, and end events. 

```js
Kefir.fromNodeStream = function (stream) {
  return Kefir.stream(function (emitter) {
    stream.on("data", emitter.emit)
    stream.on("error", emitter.error)
    stream.on("end", emitter.end)
  }
}
```

I'm lifting node database streams in a project, so maybe others want it too?

rpominov:
I didn't work with node streams yet (I do frontend mostly), but as far as I understand they are a bit different from *event* streams, because they represent stream *of data*. I not yet sure how well data streams and event streams fit together.

Rx has implementation of similar method (methods?) in https://github.com/Reactive-Extensions/rx-node , perhaps we could borrow ideas from there.

aj0strow:
Yeah it only makes sense for object mode: 

https://nodejs.org/api/stream.html#stream_object_mode

For example lots of libs have object streams, like knex for postgres.

http://knexjs.org/#Streams-stream

It's very similar to `fromEvents` helper but it would map multiple events at once. What do you think of a generic event emitter helper, and then maybe a readable stream helper?

```js
// interop for node event emitter

Kefir.fromEventEmitter(emitter, { value: "data", end: "end" })
observable.toEventEmitter({ value: "data", end: "finish" })

// maybe node stream helper

Kefir.fromReadableStream(nodeStream)
```

Or maybe just one or the other?

hallettj:
I was just typing up a test case before I found this issue. For my use, the `fromEvents` would work nicely for streams - except that the Kefir stream does not end when the Node stream ends.

```js
import * as Kefir from 'kefir'
import { Readable } from 'stream'

const nodeStream = new Readable()
const kefirStream = Kefir.fromEvents(nodeStream, 'data')

kefirStream.onValue(data => console.log('data', data))
kefirStream.onEnd(() => console.log('end (kefir callback)')) // <-- never runs

nodeStream.on('end', () => console.log('end (node callback)')) // <-- does run

nodeStream.push('some data', 'utf8')
nodeStream.push(null)  // signals end of stream
```

[...]

PR #179, opened 2016-01-14
Readable stream

aj0strow:
In reference to #178. I realized when putting this together i forgot the deactivation case in my project. 

Yeah, if it looks helpful, feel free to merge. If it's out of scope, people can see in the issue search results. Thanks for the wonderful package.


rpominov:
This probably will have to hang for a while as I'm not sure yet if it a good idea to add this. But thank you for putting this together, it's always good to have some strawman implementation especially with tests.

hastebrot:
I ran into some issues with Readable streams and Kefir streams. The buffer/string from `stdout` and `stderr` from `child_process.spawn()`/`child_process.execFile()` arrives in wrong order when subscribing to and processing events.

PR #183, opened 2016-02-15
Give access to the source code in npm releases

askmatey:
To reduce the overhead of my application, I want to access the source code directly like

```js
import Kefir from 'kefir/src/index'
```

Webpack is bloat for me.

Minor janitorial exercises are included in this commit
See https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package


rpominov:
I was thinking about it too. But decided not to support that, because we might want to rearrange files in future. We could merge this, and you'll have to rely on current files arrangement which may change in any version in future. Not sure it's a good decision though.

rpominov:
Ah, also files in src is in es6, so you may need to transpile them with babel. If we want to do it properly we should add transpiled but not bundled files to npm package. But again the files arrangement will became part of public API that we'll need to keep backward compatible, which I not interested in supporting at the moment.

askmatey:
I would even suggest, to map the `main` property in package.json to `src/index.js`. At this moment, the library comes pre-compiled out of the box. This is sub-optimal. Developers should be able to decide, how their source code should be transpiled at the end. Besides that, Webpack gives me a slight overhead to my application, since I prefer to use Browserify. 

To transform packages within `node_modules` is to add

```json
{
  "browserify": {
    "transform": ["babelify"]
  }
}
```

to package.json.

You could still include the pre-compiled builds in `/dist` to the releases.

rpominov:
> Besides that, Webpack gives me a slight overhead to my application, since I prefer to use Browserify.

Hm, you don't need to use Webpack if you want to use kefir in an app that is bundled using Browserify.

Now I'm not sure I understood you right. Could you explain again what you want to be able to do with kefir npm package in your app?

rpominov:
> Developers should be able to decide, how their source code should be transpiled at the end.

This is not _their_ source code, this is _our_ source code :)
They should only decide how their app code transpiled in my opinion.

askmatey:
Some developers may want to transpile *the* code to a ES5/ES6 hybrid. Lexical declarations are supported by all modern browsers for example. Besides that, Webpack adds a wrapper around Kefir. This is bloat for everybody using alternative bundlers.

rpominov:
> Besides that, Webpack adds a wrapper around Kefir. This is bloat for everybody using alternative bundlers.

Ah, this is just couple of bytes, I wouldn't consider it a problem.

> Some developers may want to transpile the code to a ES5/ES6 hybrid.

Not sure about this too, we may want to use non standard JS extensions. Or even when we use standard ones like `for..of` we may want to be sure that they are transpiled in loose mode for better performance.

rvikmanis:
Hi there,

I'd like to suggest a better way:

Instead of pointing `main` to the source, add a `jsnext:main` entry and point that to the source:

```js
{
...
  "main": "dist/kefir.js",
  "jsnext:main": "src/index.js"
...
}
```

Newer tools, like Rollup, will use it when needed, and legacy tools will continue to work as usual.

https://github.com/rollup/rollup/wiki/jsnext:main#we-need-your-help

rpominov:
`jsnext:main` sounds better. But rollup's tree shaking won't work for us anyway, since we add everything to `Kefir` or `Kefir.Observable.prototype` and then just export `Kefir`. Also we `require` internally instead of `import`.

Adding `jsnext:main` for rollup may make sense if we also do:
 - replace `require` with `import`
 - export all methods separately as well as complete `Kefir` object (not sure tree shaking will detect what we are doing though and work as we expect)

Anyone want to try this?

rvikmanis:
Sounds fun. I could do it probably next week.

Also, could you explain what you mean by exporting the complete object?

rpominov:
That would be great, @rvikmanis ! "exporting the complete object" is what we do now.

rvikmanis:
@rpominov do you mean this @ `src/index.js#L2`?
```js
Kefir.Kefir = Kefir;
```

rpominov:
No, `Kefir.Kefir = Kefir` is just to make sure import will work in all cases (e.g. `import {Kefir} from 'kefir'` and `import Kefir from 'kefir'`)

I mean we're currently adding everything to `Kefir` and then exporting `Kefir`. We should continue to do that, but additionally export independently what we add to `Kefir`.

For example additionally to:

```js
const map = require('./one-source/map');
Observable.prototype.map = function(fn) {
  return map(this, fn);
};
```

we should also do:

```js
export {map};
```

So user could then import only `import {constant, map} from 'kefir'` and do something like `const myProperty = map(constant(1), x => x + 2)`. In this case tree shaking hopefully will include in final bundle only parts of Kefir needed for `constant` and `map`.

rvikmanis:
Gotcha. Tree shaking should work fine with this setup.

PR #184, opened 2016-02-24
Convert source from CommonJS to ES6 module format

rvikmanis:
* Updated eslint@2.2.0 for ES module syntax.
  * Updated babel@6
  * Replaced webpack with rollup


rpominov:
> Build size increased considerably (+%22).

Hm, maybe enabling loose mode for `import/export` in babel could help, or [this](https://github.com/babel/babel-loader#babel-is-injecting-helpers-into-each-file-and-bloating-my-code) (we use babel5 thought, but it this feature available in 5 too)

There is a bigger concern though, I'll show with an example:

```js
// (supposedly we use Rollup to build this)

import {later} from 'kefir'

const stream = later(1, 'foo')

// What is the value of `.map` here? Is it a function or undefinded?
stream.map
```

Unfortunately both answers to the above questions is bad for us. If it's `undefined` (rollup didn't include transformations code) then we simply have broke the API. And if it a function then rollup basically don't work in our case (when a lot of code lives in `.prototype`), the most it can do is to exclude some creator functions from bundle (code from `primary`, `time-based`, and `interop`). Although maybe ability of excluding at least that code is valuable for us, not sure actually.

What would be cool, is to be able to include in bundle only some transformations (e.g. include `map`, but not `flatMap`), but this is probably not possible with only Rollup, we would need to change API — make those methods "static" (remove them from prototype).

rvikmanis:
I tried loose mode, but it had almost no difference.

> ...
> And if it's a function then rollup basically don't work in our case (when a lot of code lives in .prototype)

This is the case: rollup can shake the dependency tree to eliminate unused parts from the final build, but it can't split JS objects by determining which attributes get accessed; if something gets exported (like `Kefir.Observable`, for example), and is used by other parts of the exported code, every part of it stays in the build, even if we don't really need most (or any) of it on its own, because rollup has no way of knowing that.

Still rollup does a better job than webpack when bundle source consists of ES modules. Output is very clean and minimal.

**webpack + CommonJS**
```
112K   dist/kefir.js  
52K    dist/kefir.min.js
```

**webpack + ES modules**
```
136K   dist/kefir.js
64K    dist/kefir.min.js
```

**rollup + ES modules**
```
84K    dist/kefir.js
44K    dist/kefir.min.js
```

rpominov:
Wow, I didn't thought about using Rollup for building our bundle, this looks great!

rvikmanis:
> What would be cool, is to be able to include in bundle only some transformations (e.g. include map, but not flatMap), but this is probably not possible with Rollup alone, we would need to change API — make those methods "static" (remove them from prototype).

That'd be very useful. :+1:  Do you have a roadmap?

package.json
```diff
@@ -30,22 +30,31 @@
   "license": "MIT",
   "devDependencies": {
     "@reactivex/rxjs": "5.0.0-alpha.1",
-    "babel-core": "5.8.23",
+    "babel-core": "^6.5.2",
```
rpominov: Please use exact versions for all dependencies.

rpominov:
No, this is just an blurry idea, I didn't thought too much about it. One of my concerns is that it will make API surface area bigger which means more docs/test/backward compatibility/etc to maintain. Not really sure if I want to do all that :)

It was discussed a bit here https://github.com/rpominov/kefir/issues/84

rpominov:
So about this PR, I would gladly merge it when it ready. Couple of things we should make sure still work:
 - all tests running normally and pass
 - we still produce an [UMD](https://github.com/umdjs/umd) module in our bundle

rvikmanis:
Should webpack be removed from dev dependencies?

rpominov:
Yep, if it's not used anymore. Or do we still need it for something?

rvikmanis:
> all tests running normally and pass  

They do: https://travis-ci.org/rpominov/kefir/jobs/111593315


> we still produce an UMD module in our bundle  

We do: https://github.com/rvikmanis/kefir/commit/27ebf2f9f05a5851523d338513b313574f1b6b96#diff-a35864fcc4749a3b904641b4be9feacfR36

-----

One more thing,

Now that Kefir will use ES6 modules, could we distribute *src/* in the NPM package and have `jsnext:main` pointed to src/index.js?

rpominov:
Great! I'll take a closer look a bit later.

> Now that Kefir will use ES6 modules, could we distribute src/ in the NPM package and have jsnext:main pointed to src/index.js?

Yeah, don't see why not.

rpominov:
Thanks @rvikmanis, very appreciate your work!

I'll release a new version with this soon.

rpominov:
Done, `3.2.1` is out.

PR #200, opened 2016-04-26
fix(Symbol.observable): Ensure compatability with other polyfills

### LINKED ISSUE #199: Current Symbol.observable implementation might not be compitable with other polyfills
I recommend using [symbol-observable](http://npmjs.org/packages/symbol-observable). It's what's used in RxJS, Redux, and soon to be zen-observable.

The problem I see in the implementation here, is that you don't want to create the symbol with `Symbol.for`.. That was a gaff in earlier polyfills of `Symbol.observable`. 

The reason it's a gaff is prior art: `Symbol.iterator !== Symbol.for('iterator'); // true`

blesh:
`Symbol.observable !== Symbol.for('observable')`

Moves to using [symbol-observable](http://npmjs.org/package/symbol-observable)

Resolves #199 

With 💖 ... the RxJS community. 🍻

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,5 +60,8 @@
     "transducers-js": "0.4.174",
     "transducers.js": "0.3.2",
     "zen-observable": "0.1.3"
+  },
+  "dependencies": {
+    "symbol-observable": "^0.2.4"
   }
 }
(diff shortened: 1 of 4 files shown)


rpominov:
Thanks for the heads-up and for the PR! I'll take a closer look soon.

rpominov:
I've released `3.2.2` that includes this fix. Thanks again, @blesh 👍

PR #209, opened 2016-06-29
Add observable symbol to interop prototype

mAAdhaTTah:
The current implementation doesn't include the symbol on the
prototype, meaning it can't be discovered as an observable by
checking `obs[$$observable]`. This rectifies that issue.

### DIFF
--- a/src/interop/to-es-observable.js
+++ b/src/interop/to-es-observable.js
@@ -1,3 +1,4 @@
+import $$observable from 'symbol-observable';
 import {extend} from '../utils/objects';
 import {VALUE, ERROR, END} from '../constants';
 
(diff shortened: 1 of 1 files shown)


rpominov:
Thanks! I'll release a new version later this week.

mAAdhaTTah:
Great, thank you! Now I can remove all those `|| typeof ob.subscribe === 'function'` checks :).

rpominov:
@mAAdhaTTah Just released `3.2.4`, could you check if it works? I've also updated `symbol-observable` so something could go wrong.

mAAdhaTTah:
Blech, I don't think this worked. Because Symbols aren't enumerable, they don't get extended onto the prototype when calling `extend` like that. It worked in my test cuz I was running it in a PhantomJS environment, where it falls back to the string `@@oberservable` (since it doesn't have `Symbol`), which is an `ownProp`.

It has to be defined on the prototype directly. My apologies for the bum PR, will open a new one later today.

rpominov:
Ok, no problem 👌

PR #210, opened 2016-07-04
Assign $$observable to prototype directly

mAAdhaTTah:
The previous implementation didn't work because you can't assign
an object's `Symbol`s to another object. Symbol properties
aren't enumerable, so they don't get assigned with `extend`.
It has to be assigned directly to the prototype object.

### DIFF
--- a/src/interop/to-es-observable.js
+++ b/src/interop/to-es-observable.js
@@ -20,12 +20,13 @@ extend(ESObservable.prototype, {
 
     this._observable.onAny(fn);
     return () => this._observable.offAny(fn);
-  },
-  [$$observable]() {
-    return this;
   }
 });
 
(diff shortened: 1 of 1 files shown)


src/interop/to-es-observable.js
```diff
@@ -20,12 +20,12 @@ extend(ESObservable.prototype, {
 
     this._observable.onAny(fn);
     return () => this._observable.offAny(fn);
-  },
-  [$$observable]() {
-    return this;
   }
 });
 
+ESObservable.prototype[$$observable] = function() {
```
rpominov: A comment explaining why it's done directly would be helpful here.

mAAdhaTTah:
Done. There's a fuller explanation in the git commit as well.

rpominov:
Thanks! Released `3.2.5` with it.

PR #211, opened 2016-07-05
Add observe method to Kefir.Observable

### LINKED ISSUE #151: Simplify subscription methods?
Extracting this from #103 to own issue, see https://github.com/rpominov/kefir/issues/103#issuecomment-100713065 and other comments on that issue.

mAAdhaTTah:
I'm going to lead off by saying I know this doesn't match the [comment](https://github.com/rpominov/kefir/issues/103#issuecomment-100713065) you made. I don't have strong feelings about the difference between `observe` / `subscribe` and the subscription token vs unsub function. However, I do think if we're going to add a method that does something similar to `subscribe`, we should adhere to standards for better interoperability. That's what this matches up with.

I'm happy to change around the layout if you'd really prefer the alternate you laid out above, or even include them both, but I'd like to make the case that adhering to the es-observable standard is the way to go here.

I'm still getting a feel for the codebase, so if there's a better way to implement this, definitely interested in feedback as well.

Fixes #151.

### DIFF
--- a/docs-src/descriptions/main-methods.jade
+++ b/docs-src/descriptions/main-methods.jade
@@ -1,5 +1,57 @@
 h2#main-methods Subscribe / add side effects
 
++descr-method('observe', 'observe', 'obs.observe(Observer|callback)').
+  Subscribes the provided Observer to #[tt obs]. Observer is an object with 3 optional methods:
+
+ul
+  li #[b next] - called on values from the observable
+  li #[b error] - called on errors from the observable
+  li #[b complete] - called on end of the observable
+
+p.
+  Returns a #[tt Subscription] object, which has
+  an #[tt unsubscribe] method and a read-only #[tt closed] property.
+  #[tt closed] indicates whether the #[tt unsubscribe] method has been called
+  and the subscription to the observable has been closed.
+
+pre.javascript(title='example')
+  :escapehtml
+    var stream = Kefir.sequentially(1000, [1, 2]);
+    subscription = stream.observer({
+      next(value) {
+        console.log('value', value)
+      },
+      error(value) {
+        console.log('error', value)
+      },
+      complete(value) {
+        console.log('closed', subscription)
+      },
+    });
+
+pre(title='console output')
+  :escapehtml
+    > value: 1
+    > value: 2
+    > complete: undefined
+
+p.
+  In addition to passing in an Observer, #[tt observe] can takes 1-3 callbacks individually:
+
+pre.javascript(title='example')
+  :escapehtml
+    var stream = Kefir.sequentially(1000, [1, 2]);
+    stream.observer(
+      function onNext(value) {
+        console.log('value', value)
+      },
+      function onError(value) {
+        console.log('error', value)
+      },
+      function onComplete(value) {
+        console.log('closed', subscription)
+      }
+    );
 
 +descr-method('on-value', 'onValue', 'obs.onValue(callback)').
   Subscribes #[b callback] to values on an observable.
--- a/docs-src/descriptions/main-methods.jade
+++ b/docs-src/descriptions/main-methods.jade
@@ -17,7 +17,7 @@ p.
 pre.javascript(title='example')
   :escapehtml
     var stream = Kefir.sequentially(1000, [1, 2]);
-    subscription = stream.observer({
+    var subscription = stream.observer({
       next(value) {
         console.log('value', value)
       },
@@ -28,6 +28,10 @@ pre.javascript(title='example')
         console.log('closed', subscription)
       },
     });
+
+    ...
+    // later
+    subscription.unsubscribe()
 
 pre(title='console output')
   :escapehtml
--- a/src/observable.js
+++ b/src/observable.js
@@ -126,33 +126,25 @@ extend(Observable.prototype, {
       };
     }
 
-    if (observer.next) {
-      this.onValue(observer.next);
-    }
+    const fn = function(event) {
+      if (event.type === VALUE && observer.next) {
+        observer.next(event.value);
+      } else if (event.type === ERROR && observer.error) {
+        observer.error(event.value);
+      } else if (event.type === END && observer.complete) {
+        observer.complete(event.value);
+      }
+    };
 
-    if (observer.error) {
-      this.onError(observer.error);
-    }
-
-    if (observer.end) {
-      this.onEnd(observer.complete);
-    }
+    this.onAny(fn);
 
     return {
       unsubscribe() {
-        if (observer.next) {
-          _this.offValue(observer.next);
-        }
+        if (!closed) {
+          _this.offAny(fn);
 
-        if (observer.error) {
-          _this.offError(observer.error);
+          closed = true;
         }
-
-        if (observer.end) {
-          _this.offEnd(observer.complete);
-        }
-
-        closed = true;
       },
       closed() {
         return closed;
--- a/src/observable.js
+++ b/src/observable.js
@@ -133,6 +133,7 @@ extend(Observable.prototype, {
         observer.error(event.value);
       } else if (event.type === END && observer.complete) {
         observer.complete(event.value);
+        closed = true;
       }
     };
 
--- a/src/observable.js
+++ b/src/observable.js
@@ -114,15 +114,15 @@ extend(Observable.prototype, {
     return this._off(ANY, fn);
   },
 
-  subscribe(observer) {
+  subscribe(observer, onError, onComplete) {
     const _this = this;
     let closed = false;
 
     if (typeof observer === 'function') {
       observer = {
         next: observer,
-        error: arguments[1],
-        complete: arguments[2]
+        error: onError,
+        complete: onComplete
       };
     }
 
--- a/src/observable.js
+++ b/src/observable.js
@@ -114,6 +114,52 @@ extend(Observable.prototype, {
     return this._off(ANY, fn);
   },
 
+  subscribe(observer) {
+    const _this = this;
+    let closed = false;
+
+    if (typeof observer === 'function') {
+      observer = {
+        next: observer,
+        error: arguments[1],
+        complete: arguments[2]
+      };
+    }
+
+    if (observer.next) {
+      this.onValue(observer.next);
+    }
+
+    if (observer.error) {
+      this.onError(observer.error);
+    }
+
+    if (observer.end) {
+      this.onEnd(observer.complete);
+    }
+
+    return {
+      unsubscribe() {
+        if (observer.next) {
+          _this.offValue(observer.next);
+        }
+
+        if (observer.error) {
+          _this.offError(observer.error);
+        }
+
+        if (observer.end) {
+          _this.offEnd(observer.complete);
+        }
+
+        closed = true;
+      },
+      closed() {
+        return closed;
+      }
+    };
+  },
+
   // A and B must be subclasses of Stream and Property (order doesn't matter)
   _ofSameType(A, B) {
     return A.prototype.getType() === this.getType() ? A : B;
--- a/src/observable.js
+++ b/src/observable.js
@@ -114,7 +114,7 @@ extend(Observable.prototype, {
     return this._off(ANY, fn);
   },
 
-  subscribe(observer, onError, onComplete) {
+  observe(observer, onError, onComplete) {
     const _this = this;
     let closed = false;
 
@@ -146,7 +146,7 @@ extend(Observable.prototype, {
           closed = true;
         }
       },
-      closed() {
+      get closed() {
         return closed;
       }
     };
--- a/test/specs/kefir-observable.js
+++ b/test/specs/kefir-observable.js
@@ -1,0 +1,51 @@
+let {Kefir} = require('../test-helpers.coffee');
+
+describe('Kefir.Observable', function () {
+  describe('observe', function () {
+    let em, count, obs, sub;
+
+    beforeEach(function () {
+      em = null;
+      count = 0;
+      obs = Kefir.stream(_em => {
+        em = _em;
+      });
+      sub = obs.observe({next: () => count++, error: () => count--, complete: () => count = 0});
+    });
+
+    it('should return a Subscription', function () {
+      expect(sub.closed).toBe(false);
+      expect(typeof sub.unsubscribe).toBe('function');
+    });
+
+    it('should call Observer methods', function () {
+      expect(count).toBe(0);
+
+      em.emit(1);
+      expect(count).toBe(1);
+
+      em.emit(1);
+      expect(count).toBe(2);
+
+      em.error(1);
+      expect(count).toBe(1);
+
+      em.end();
+      expect(count).toBe(0);
+      expect(sub.closed).toBe(true);
+    });
+
+    it('should unsubcribe early', function () {
+      expect(count).toBe(0);
+
+      em.emit(1);
+      expect(count).toBe(1);
+
+      sub.unsubscribe();
+
+      em.emit(1);
+      expect(count).toBe(1);
+      expect(sub.closed).toBe(true);
+    });
+  });
+});
--- a/test/specs/kefir-observable.js
+++ b/test/specs/kefir-observable.js
@@ -1,10 +1,10 @@
 let {Kefir} = require('../test-helpers.coffee');
 
-describe('Kefir.Observable', function () {
-  describe('observe', function () {
+describe('Kefir.Observable', () => {
+  describe('observe', () => {
     let em, count, obs, sub;
 
-    beforeEach(function () {
+    beforeEach(() => {
       em = null;
       count = 0;
       obs = Kefir.stream(_em => {
@@ -13,12 +13,12 @@ describe('Kefir.Observable', function () {
       sub = obs.observe({next: () => count++, error: () => count--, complete: () => count = 0});
     });
 
-    it('should return a Subscription', function () {
+    it('should return a Subscription', () => {
       expect(sub.closed).toBe(false);
       expect(typeof sub.unsubscribe).toBe('function');
     });
 
-    it('should call Observer methods', function () {
+    it('should call Observer methods', () => {
       expect(count).toBe(0);
 
       em.emit(1);
@@ -35,7 +35,7 @@ describe('Kefir.Observable', function () {
       expect(sub.closed).toBe(true);
     });
 
-    it('should unsubcribe early', function () {
+    it('should unsubcribe early', () => {
       expect(count).toBe(0);
 
       em.emit(1);
--- a/docs-src/includes/side-menu.jade
+++ b/docs-src/includes/side-menu.jade
@@ -32,6 +32,7 @@ ul.toc-section
 
 a.toc-title(href='#main-methods') Subscribe / add side effects
 ul.toc-section
+  +method('#observe', 'observe')
   +method('#on-value', 'onValue')
   +method('#off-value', 'offValue')
   +method('#on-error', 'onError')
--- a/src/emitter.js
+++ b/src/emitter.js
@@ -20,5 +20,5 @@ export default function emitter(obs) {
     return obs._active;
   }
 
-  return {value, error, end, event, emit: value, emitEvent: event};
+  return {value, error, end, event, emit: value, emitEvent: event, next: value, complete: end};
 }


src/observable.js
```diff
@@ -114,6 +114,52 @@ extend(Observable.prototype, {
...
   },
 
+  subscribe(observer) {
+    const _this = this;
+    let closed = false;
+
+    if (typeof observer === 'function') {
+      observer = {
+        next: observer,
+        error: arguments[1],
```
rpominov: Not sure we need to support `subscribe(onValue, onError)` format. Supporting `subscribe({onValue, onError})` + `subscribe(onValue)` is enough IMO. Is this format from ES spec?

src/observable.js
```diff
@@ -114,6 +114,52 @@ extend(Observable.prototype, {
...
+        complete: arguments[2]
+      };
+    }
+
+    if (observer.next) {
+      this.onValue(observer.next);
```
rpominov: I think it's best to use `this.onEvent()` under the hood of `subscribe()` similar to how it done in https://github.com/rpominov/kefir/blob/master/src/interop/to-es-observable.js

src/observable.js
mAAdhaTTah: Yeah, in the [es-observable](https://github.com/zenparsing/es-observable) spec, under [Observable](https://github.com/zenparsing/es-observable#observable).

src/observable.js
```diff
@@ -114,6 +114,52 @@ extend(Observable.prototype, {
...
+
+    if (observer.error) {
+      this.onError(observer.error);
+    }
+
+    if (observer.end) {
+      this.onEnd(observer.complete);
+    }
+
+    return {
```
rpominov: That's strange. In https://github.com/rpominov/kefir/blob/master/src/interop/to-es-observable.js we return function from `.subscribe()` not an object. Either here or there we have an incompatibility with the spec.

src/observable.js
rpominov: Ah, it's fine then. I'd only use explicit arguments instead of `arguments[1/2]`: 

```js
subscribe(observer, onError, onComplete) { ...
```

rpominov:
Agree with making it compatible with standard. Did you see my comment here btw https://github.com/rpominov/kefir/issues/151 (sorry, forgot to add a mention) ?

Also we need tests and docs. Can you add it?

src/observable.js
mAAdhaTTah: I can definitely port over that code. Here's the thing: If we're going to add a `subscribe` function that matches the `es-observable` spec, then we don't actually need to provide the `toESObservable` method. Should we remove/deprecate that method are part of this PR?

I ask because we'd end up with code duplication if I just copied it.

src/observable.js
mAAdhaTTah: Based on the spec, it appears it's supposed to return

src/observable.js
rpominov: > then we don't actually need to provide the toESObservable method.

Not exactly, `toESObservable` also adds `.takeErrors(1)` so the stream would end on the first error as spec requires. So we probably still need it as 100% compatible es-observable.

As for duplication, we probably could do something like `Observable.prototype.subscribe = ESObservable.prototype.subscribe` if they end up completely identical. Otherwise I'm ok with a bit of duplication.

src/observable.js
rpominov: I see. Would be cool to make it one method (O.prototype.subscribe === ESO.prototype.subscribe) and fix it in one place then.

src/observable.js
mAAdhaTTah: Ah, I see. Are there concerns about having two methods with the same name with slightly different behaviors? Although we don't have the `Symbol.observable` on the `Observable` prototype, so maybe that's mitigating?

rpominov:
Yeah sorry, about coffeescript. I want to migrate tests to JS, but can't find a time. It's ok to write new tests in JS though. I think it is already supported by the dev scripts, you'll only need to add a file like `test/specs/observable.js` with the tests and the scripts should pick it up just fine hopefully. Or if that won't work, but it'll still be easier for you to make JS files in tests work and write tests in JS, that would be great too.

Sorry about overall dev experience btw, I know it's far from perfect. I wish I had more free time to upgrade it, or someone would do it.

src/observable.js
rpominov: Yeah, I'm not a fun of this situation too. This is why I didn't want to use same name for the method originally. But I think this is still a reasonable trade of (with a note about different behavior in the docs + not having `Symbol.observable` on `Observable`).

mAAdhaTTah:
No worries. I'll continue work on the PR this evening.

> Yeah, I'm not a fun of this situation too. This is why I didn't want to use same name for the method originally. But I think this is still a reasonable trade of (with a note about different behavior in the docs + not having Symbol.observable on Observable).

Would it be really weird to have 2 methods: a `subscribe` method that adheres exactly to the `es-observable` spec (e.g. ends on error) and an `observe` method with different semantics and behavior, adhering to Kefir's current behavior? Then we could add the `Symbol.observable` to Kefir's Observable prototype and eliminate the `toESObservable` method, and (maybe?) deprecate the `onValue` etc. methods.

Related to all of these questions, how much do you want to change Kefir to match the spec? I would obviously push to make it match as much as possible, given the importance standards have been for making JavaScript reasonable to work with, but I understand there may be good reasons to avoid full compliance (especially if it requires a breaking change like stream behavior).

rpominov:
We probably don't want too much changes in Kefir. And maybe we should name it `.observe` after all. If we add `.subscribe` to `Kefir.Obsevable` we probably also should add `Observable.of` and `Observable.from`, also later we'll need to add / make compatible [all these methods](https://github.com/zenparsing/zen-observable#extended-api) because they are going to be in the spec. This is just too much work, and also ES spec still a moving target, so we'll need to update all this while spec changes. Not even mentioning semantics changes (end on first error etc.).

Considering all this, it's probably best to only make it possible to do `Observable.from(myKefirObservable)` — make Kefir observables convertible to ES observables. This is already done since we have `obs[Symbol.observable]()`, but we need to fix it a bit (return an object form `subscribe`).

So I propose: 1) leave compatibility level as it is now, 2) add a new method called `observe` which will be similar to `subscribe` but not necessarily compatible with ES spec, 3) we don't take `Kefir.Observable.subscribe` name now, and will be able to go full compatibility road in the future. What do you think?

mAAdhaTTah:
FWIW, the "extended API" linked is not part of the spec and isn't going to be in its first version, so I don't necessarily think there's an issue there. I think they're trying to get a simple version accepted as part of the language so various implementations have interop targets.

However, I do think the "spec as a moving target" could be an issue, since it's still being hashed out, and more importantly, that the semantics of `subscribe` would have to change to match. It would be weird to have a `subscribe` method that doesn't conform with the spec in that way.

> This is already done since we have `obs[Symbol.observable]()`, but we need to fix it a bit (return an object from `subscribe`).

Looks like you fixed this already :). I would have `observe` return a similar object, so we can get as close to the spec as possible and make future transitions easier.

So I think I agree with you, given the divergence so far between the current, early spec and Kefir. I will implement `observe` as close as we can get to the spec, though, to simplify the transition for users later.

mAAdhaTTah:
Now with docs & tests.

docs-src/descriptions/main-methods.jade
```diff
@@ -1,5 +1,57 @@
...
+      next(value) {
+        console.log('value', value)
+      },
+      error(value) {
+        console.log('error', value)
+      },
+      complete(value) {
+        console.log('closed', subscription)
+      },
+    });
```
rpominov: Let's add 

```
...
// later
subscription.unsubscribe()
```
here.

docs-src/descriptions/main-methods.jade
```diff
@@ -1,5 +1,57 @@
...
+p.
+  Returns a #[tt Subscription] object, which has
+  an #[tt unsubscribe] method and a read-only #[tt closed] property.
+  #[tt closed] indicates whether the #[tt unsubscribe] method has been called
+  and the subscription to the observable has been closed.
+
+pre.javascript(title='example')
+  :escapehtml
+    var stream = Kefir.sequentially(1000, [1, 2]);
+    subscription = stream.observer({
```
rpominov: var

src/observable.js
```diff
@@ -114,6 +114,45 @@ extend(Observable.prototype, {
...
+      unsubscribe() {
+        if (!closed) {
+          _this.offAny(fn);
+
+          closed = true;
+        }
+      },
+      get closed() {
```
rpominov: Are getters supported by all browsers?

test/specs/kefir-observable.js
```diff
@@ -0,0 +1,49 @@
+let {Kefir} = require('../test-helpers.coffee');
+
+describe('Kefir.Observable', function() {
+    describe('observe', function() {
```
rpominov: Let's use two spaces for consistency with the rest of JS here.

test/specs/kefir-observable.js
```diff
@@ -0,0 +1,49 @@
...
+        let em, count, obs, sub;
```
rpominov: Not sure if we can use es6 here. I mean I'm not sure it's transpiled, so if we use es6 then tests may work only in modern environments.

rpominov: Would be cool to make sure it's transpiled and keep using es6.

test/specs/kefir-observable.js
```diff
@@ -0,0 +1,49 @@
...
+
+        beforeEach(function() {
```
rpominov: If we use es6, I'd use arrow functions whenever possible.

src/observable.js
mAAdhaTTah: [IE9+](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get). Do you need IE8 support?

test/specs/kefir-observable.js
mAAdhaTTah: Sure, so used to mocha where you're not supposed to.

test/specs/kefir-observable.js
mAAdhaTTah: I can look into transpiling it, but are you running the tests in a browser or node? Because node has had support for `let` and arrow functions since [Node 4](http://node.green/).

rpominov: We want to be able to run them in browsers. We have that link in readme: 
image

Ability to run test in older node versions is also nice, in case we'll want to check if we still support them.

src/observable.js
rpominov: Is there an easy way to do some kind of fallback in IE8? If it's too much trouble we can drop IE8 support, but if adding fallback is easy would be nice to keep it.

mAAdhaTTah: The only fallback would be to just not use a getter, I think.

rpominov: Not use at all, or only in IE8?

test/specs/kefir-observable.js
mAAdhaTTah: Some of those tests fail currently, although I'm not sure why. I'll look into transpiling the tests in Jasmine.

src/observable.js
mAAdhaTTah: If we want IE8 support, we probably shouldn't use a getter.

test/specs/kefir-observable.js
rpominov: Yeah, async tests fail in browser currently. Sorry about that. This is needs to be fixed, but for now you can just ignore them.

rpominov: See https://github.com/rpominov/kefir/issues/213

src/observable.js
rpominov: I see, let's drop it then.

rpominov:
Overall it looks good now. I will take a closer look and merge at the end of the week. 

Also if there anything you want to add/change before we merge please let me know, so I don't merge before it's ready.

rpominov:
Ah, forgot about es6 in tests. But if there any problems, I probably will be able to do it by myself at weekend.

mAAdhaTTah:
Just ES6 in tests for this PR. I think there were a few minor things we also wanted to do for `es-observable` compliance (e.g. adding  `next` & `complete` methods to the `emitter`), but either you can grab them if you get to them first, or I'll open a separate PR for them later. Either way, they shouldn't hold up this PR.

rpominov:
> adding next & complete methods to the emitter

Right. Forgot also. If you want to do this in this PR, that would be cool.

mAAdhaTTah:
Done.

rpominov:
Regarding es6+ in test, I [added](https://github.com/rpominov/kefir/commit/fbca5f1786be02c2fea8303aa5652643dc641d44) babelify to browser bundle, but we still run untransformed code on node. Thinking of leaving it as is for now.

mAAdhaTTah:
If you bump the node minimum to 4+, a lot of these features will work natively, so it should be fine.

rpominov:
Two things still bother me about how API ended up looking:

1) We use "value" in the rest of the API for value-events (ex. `ignoreValue`, `event.type==='value'`), but in `observe` we use "next" (same for `end` vs `complete`)
2) Originally I wanted `observe` to replace all methods for subscribing, but currently it can't replace `.onEvent`.

I'm not happy with this inconsistency, `observe` looks a bit alien to the rest of API, and the API as a whole looks less solid. I'm thinking about fixing it by renaming `next -> value`, `complete -> end`, and adding `event` to subscribe object. Another way we could go is to rename them in the rest of API, but it's much harder to do. What do you think?

rpominov:
I know, I know, we keep getting back to the [original proposal](https://github.com/rpominov/kefir/issues/103#issuecomment-100713065), but that's just because I keep remembering the reasons why I've chosen those names back then.

mAAdhaTTah:
For starters, I think it does replace `onAny`. Maybe you have more experience w/ it, but I don't see a use case for `onAny` that doesn't involve checking the `event.type`, in which case, if you want to respond to both `value` and `error` events, you would just pass an Observer into `observe` with only `next` and `error` methods, so you skip the `type` checking in userland and just pass in callbacks for the events you want. It wouldn't be difficult to reuse a callback for multiple methods in the Observer, so I would not add an `event` method, were we to switch the API.

I think the first point is more valid though; we're kinda stuck between the legacy of the library itself and the `es-observable` spec. Given my preference for spec-compliance, I'd actually push to rename the rest of the API, but that would be a pretty major backwards-breaking change.

That said, I don't think it would _actually_ be that hard. Change `event.type -> value`, find a new name for `ignoreValues` (Rx calls it [ignoreElements](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/ignoreelements.md)), and I think that's it, really. Probably a bunch of documentation changes too.

I don't know as I would suggest making a BC break over semantics in the short-term, but I think it's worth considering first whether (and when) you'd want to make the library more spec-compliant, and making sure that whatever decisions we make now about the API doesn't make that more difficult. We did save the `subscribe` "namespace" for that purpose, so it may not be a big deal to change the `observe` method to be more consistent internally. Part of my goal was to make transition from `observe` to an eventual `subscribe` method easier.

So ultimately, I think it depends on a couple questions for you:

1. What are your plans for making the library spec-compliant in the future? Is including `Symbol.observable` enough or do you eventually want the library's API to be spec-compliant as well?
2. How willing are you to deprecate features and/or break BC in order make the library spec-compliant?
3. How important is making the library spec-compliant to your API design right now? Do you want an easier upgrade path for users or each major version to be internally consistent?

rpominov:
> For starters, I think it does replace onAny ...

Generally I agree, but there are some minor reasons for supporting `observe({event})`:

  1. Sometimes it's just a bit more convenient to pass single callback rather than split in 2 or 3 with partial application or something.
  2. For performance it again might better to pass single callback rather than do partial application or functions wrapping etc.
  3. Ideally I would like if `emitter` interface would match interface of the objects that we pass to `observe`, so we could do `stream.observe(emitter)`. Not sure what are use cases for this, but it's just feel right if those interfaces match.

> That said, I don't think it would actually be that hard.

Here are all places where we would need a rename (if I didn't miss anything):

```
onValue
offValue
onEnd
offEnd
event.type==='value'
event.type==='end'
ignoreValues
ignoreEnd
beforeEnd
```

Yeah, maybe not so many, also considering that first 4 will be removed eventually. Yet `event.type==='next'` seems wrong to me for example. Also in Kefir we can have many errors in a stream, so in Kefir we can say "next error" as well as "next value".

As for plans, I don't have any at the moment because I don't use Kefir in any of my projects, and it hard to me to understand how painful (or not) changes to API would be for people who maintain projects that use Kefir, and also to understand what advantages spec compliance will give to them. Also I don't know what changes are necessary for 100% compliance, and whether those changes will make sense for Kefir. Two thing I'm particularly against of is automatic exceptions catching, and completion on first error. Also I want to wait for spec to be finished before making any plans.

Why spec compliance is important to you btw?

mAAdhaTTah:
Definitely agree w/ 3. For one and two, I'm less sure of the use cases, but I haven't yet run into a case where I needed `onAny` (except to write this `observe` method). The other issue is you wouldn't be able to pass the emitter in directly, or you'd have to make sure `observe` doesn't call both `event` and `next` on the same Observer, or you'd get each event twice.

> Also in Kefir we can have many errors in a stream, so in Kefir we can say "next error" as well as "next value".

This is true, and would definitely be a major BC break to conform to the spec. However, terminating on error resolves the semantic tension between `next` and `error`, as `next` would then only refer to the next value, `error` would refer to the single error event that terminates the stream, and `complete` would refer to the single completion event that terminates the stream.

> ...it hard to me to understand how painful (or not) changes to API would be for people who maintain projects that use Kefir, and also to understand what advantages spec compliance will give to them.

I personally like how React has gone about it: aggressively deprecate features as they realize the problems that arise as a result of them, while providing good warnings and documentation for developers on how to upgrade their current code. We could roll out these changes over a number of 3.x versions and remove the deprecated code in 4.x, similar to how you did going from 2.x -> 3.x. As long as there's a clear migration path, it shouldn't be too bad for consumers.

We could also maintain both the 4.x & 3.x versions for a while, backporting changes, or release 4.x as a beta for a while, with messages suggesting the developer upgrades and tests the beta in the 3.x branch. It all depends on the amount of work we'd be willing & interested in committing to library consumers.

> I don't know what changes are necessary for 100% compliance

Here's my current read on the changes required:

1. Implement the static methods `of` & `from`.
2. Add `subscribe` method that is spec-compliant.
    * I think the `observe` method as it currently exists is fairly spec-compliant, sans `end` behavior.
3. Terminate streams on `error` by default.
4. (maybe) Rename `value` methods to `next` methods.

I think number 3 is the big sticking point for Kefir, although I don't really know what the impact of automatic exception catching would be. If it's a question of disagreeing with the principle, then you're probably better off abandoning spec-compliance, in all honesty. However, there could be a migration path for creating non-error-terminating streams in a hypothetical 4.x:

```js
// Resulting stream would not end
// if error emitted.
stream$.surviveErrors()
    .subscribe(
        x => console.log('next', x),
        err => console.log('error', err),
        () => console.log('complete')
    );
```

In the same way it's possible to create an error-teminating stream now with `takeErrors(1)`, we could do the reverse in 4.x to allow the developer to get a non-terminating stream.

> Why spec compliance is important to you btw?

Couple reasons: The most direct reason is I'm working on a reactive framework underpinned by Kefir. However, I'd really like to be able to say the framework returns an `Observable` rather than a `Kefir.Observable`. That would make it easier for consumers of the library to understand how it works, rather than learning how a particular implementation works.

It would also mean the consumer of the library wouldn't have to worry about which implementation we're using, which would give us more flexibility down the line, should we need to change implementations or upgrade versions. In the long future, if there's better library interop, it would be ideal to allow the consumer to swap out which reactive library it uses, so they wouldn't have to pull in two reactive libraries, should they want to use the framework with RxJS.

And that's really the reason you want some level of interop. As an example, once Promises/A+ got standardized, it becomes super easy for you to use Bluebird as your Promise library of choice and interop with Q or native Promises returned from other libraries. In a future where `Observable` is built into the language, being able to convert to and from Observable implementations becomes important to making the ecosystem work coherently together, while giving consumers the flexibility to choose a library that most suits them.

Some of this is muted by the `Symbol.observable` interop target, but the general point is that specs are good for the ecosystem. In this case, it's a pretty basic implementation, plus its semantics, but obviously the most crucial to get right.

All of that said, my primary purpose in bringing this PR was to get the `Subscription` object pattern. It makes more sense semantically to have an object who's purpose is to unsubscribe than to hold onto the function for that purpose. For now, I'm happy to just say the framework returns a `Kefir.Observable`, since it's still in early development (building it along with a project, so I will have to extract it first), but it would be a long goal to eventually return interop targets.

That's why I'm pushing interop. There's always a possibility the framework doesn't get off the ground the way I'm planning, so it's up to you to decide the relative importance of this for Kefir as well as the timeline for doing so.

In the end, I would be comfortable with changing `observe` to be internally consistent with the API (although if it were up to me, I wouldn't include `event` as a callback), even if it's less compliant, since the timeline for the spec to be finalized is still likely a ways out. If spec compliance is something you want to pursue, I might suggest opening this discussion up in a separate issue to gather feedback from other library users and start planning what the path to spec compliance might look like. If it isn't, that wouldn't stop me from using Kefir for this framework because the API is really clear and solid and with `observe`, irons out the only area I had issues with (the `on/off` functions), so it's up to you what you want to do.

mAAdhaTTah:
I know I wrote a novel, so tl;dr version: Make the `observe` method more in line with the rest of the API, start considering `es-observable` compatibility and the possible BC-breaks that would incur.

rpominov:
Yeah, I read everything, just wanted to think about it for a bit. Since we won't move on full compatibility road at this step, I think it's good idea to keep API consistent after each change, so we should use `value` and `end` names.

As for `observe({event})` let's not add it. We can always add it later, but it will be harder to remove it later. There probably no good use cases for it indeed. And without it transition from `observe` to `subscribe` will be just matter of renaming.

I'll finish it up and release a version at weekend if nothing will change.

rpominov:
Done, `3.3.0` is out!

Issue #160, opened 2015-10-11
Fantasy Land compatibility


rpominov:
Creating this issue to investigate how we could make Kefir compatible with the [Fantasy Land Specification](https://github.com/fantasyland/fantasy-land).

I think of two possibilities: add Fantasy Land methods directly into Kefir, or create a plugin as a separate npm package that injects Fantasy Land methods into Kefir. Don't know what is better for us yet, and would like to know your opinions.

In either way first we need to find out how Fantasy Land methods map to current Kefir methods. Here is a draft of such mapping (not sure if it's correct and complete; will update it in progress):

## [Setoid](https://github.com/fantasyland/fantasy-land#setoid)

`a.equals(a)` – we can't implement this (?)

## [Semigroup](https://github.com/fantasyland/fantasy-land#semigroup)

`a.concat(b)` — `obsA.concat(obsB)` (?)

## [Monoid](https://github.com/fantasyland/fantasy-land#monoid)

`m.empty()` — `Kefir.never()` (?)

## [Functor](https://github.com/fantasyland/fantasy-land#functor)

`u.map(f)` — `obs.map(f)`

## [Apply](https://github.com/fantasyland/fantasy-land#apply)

`a.ap(b)` — can be implemented on top of `combine`

## [Applicative](https://github.com/fantasyland/fantasy-land#applicative)

`a.of(x)` — `Kefir.constant(x)` (?)

## [Foldable](https://github.com/fantasyland/fantasy-land#foldable)

`u.reduce(f, x)` — `Kefir.scan(f, x).last()` (?)

One problem that we can't return the result value directly but can only return an observable containing result value. Although FL seem to not require that.

## [Traversable](https://github.com/fantasyland/fantasy-land#traversable)

`u.sequence(of)` — not sure...

Here we have the same problem as in `reduce` — unable to return result value directly:

```
// we can't do this
Observavble (f a) -> f (Observable a)

// only can do this (based on reduce)
Observavble (f a) -> Observable (f (Observable a))
```

`Observable (f (Observable a))` already doesn't seem to be very useful, but the main problem is that we can't satisfy the following law:

> `t(u.sequence(f.of))` is equivalent to `u.map(t).sequence(g.of)` where `t` is a natural transformation from `f` to `g` (naturality)

Although the other way around form of `sequence` — `f (Observable a) -> Observavble (f a)` seems very similar to `Kefir.combine :: [Observable] -> Observable []`. Not sure what it gives us though :)

[...]

PR #215, opened 2016-08-22
Add 'chain' synonym for flatMap.

peres:
This makes operation with fantasyland-compatible libraries (e.g. Ramda) a bit smoother.

I didn't realize you had #160 open before posting this PR. I hope you can consider it anyway. Thanks.


rpominov:
I want to add full Fantasy Land compatibility, but I want to wait until https://github.com/fantasyland/fantasy-land/pull/146 lands.

Thank you for the PR, it reminded me that this is important. But if we're doing this we should add all methods not only `chain`.

PR #217, opened 2016-08-27
Add Flow type declarations

AgentME:
This pull request adds a file specifying the types of all of Kefir's functions in a way that [Flow](https://flowtype.org/) understands. Flow will be able to type-check code using Kefir with this declaration file automatically without requiring any special Kefir-specific configuration by the user. Flow does a lot of type-inference and works really well with Kefir in my experience.

Here's an example of Kefir and Flow used together with the declaration file. We have a function defined to only take streams of numbers, but we pass a stream of strings to it instead. Flow correctly detects this as a type error:

```js
/* @flow */
import Kefir from 'kefir';

function foo(stream: Kefir.Observable<number>) {
}

var s = Kefir.constant(5).map(x => String(x));
foo(s);
```

```
test.js:7
  7: var s = Kefir.constant(5).map(x => String(x));
                                        ^^^^^^^^^ function call
  7: var s = Kefir.constant(5).map(x => String(x));
                                        ^^^^^^^^^ string. This type is incompatible with
  4: function foo(stream: Kefir.Observable<number>) {
                                           ^^^^^^ number


Found 1 error
```

The declaration file was hand-written, not autogenerated or anything like that. As Kefir gets new methods, it would need to be updated by hand in order for user code using them to type-check correctly. Personally as a heavy Flow and Kefir user I don't think it's a big deal if Kefir's declaration file lags behind what Kefir supports; a Flow user can always cast values to `any` to bypass a type error when they need to do something that Flow doesn't understand. I find huge value in having any Kefir types available at all.

TODO:
- [x] The file kefir.js.flow needs to be copied into the dist/ directory (where Flow will look for it) on build. I'm not sure how best to set this up in Kefir's build process.
- [x] Optional: a test could be added that checks that the kefir.js.flow file is parseable, and verifies that Flow can typecheck a test file. We should add "flow-bin" as a devDependency if we want to do that. I could take a stab at this.

### DIFF
--- a/kefir.js.flow
+++ b/kefir.js.flow
@@ -5,7 +5,7 @@ export type Event<T> =
   {type: 'error', value: any} |
   {type: 'end', value: void};
 
-declare class _Emitter<T> {
+export type Emitter<T> = {
   value(value: T): boolean;
   event(event: Event<T>): boolean;
   error(e: any): boolean;
@@ -14,9 +14,7 @@ declare class _Emitter<T> {
   // Deprecated methods
   emit(value: T): boolean;
   emitEvent(event: Event<T>): boolean;
-}
-
-export type Emitter<T> = _Emitter<T>;
+};
 
 export type Subscription = {
   closed: boolean;
@@ -75,7 +73,7 @@ declare class Observable<+T> {
   slidingWindow(max: number, min?: number): Observable<T[]>;
   bufferWhile(predicate?: (value: T) => boolean, options?: {flushOnEnd?: boolean}): Observable<T[]>;
   transduce(transducer: any): Observable<any>;
-  withHandler<U>(handler: (emitter: _Emitter<U>, event: Event<T>) => void): Observable<U>;
+  withHandler<U>(handler: (emitter: Emitter<U>, event: Event<T>) => void): Observable<U>;
 
   combine<U>(otherObs: Observable<U>): Observable<[T,U]>;
   combine<U,Z>(otherObs: Observable<U>, combinator: (t: T, u: U) => Z): Observable<Z>;
@@ -134,11 +132,11 @@ declare var Kefir: {
   interval<T>(interval: number, value: T): Observable<T>;
   sequentially<T>(interval: number, values: T[]): Observable<T>;
   fromPoll<T>(interval: number, f: () => T): Observable<T>;
-  withInterval<T>(interval: number, f: (emitter: _Emitter<T>) => void): Observable<T>;
+  withInterval<T>(interval: number, f: (emitter: Emitter<T>) => void): Observable<T>;
   fromCallback<T>(f: (cb: (value: T) => void) => void): Observable<T>;
   fromNodeCallback<T>(f: (cb: (err: any, value: ?T) => void) => void): Observable<T>;
   fromEvents(target: Object, eventName: string, transformer?: (event: any) => any): Observable<any>;
-  stream<T>(subscribe: (emitter: _Emitter<T>) => ?() => void): Observable<T>;
+  stream<T>(subscribe: (emitter: Emitter<T>) => ?() => void): Observable<T>;
 
   constant<T>(value: T): Observable<T>;
   constantError(err: any): Observable<any>;
--- a/kefir.js.flow
+++ b/kefir.js.flow
@@ -1,19 +1,19 @@
 /* @flow */
 
-export type Event<T> =
-  {type: 'value', value: T} |
+export type Event<V> =
+  {type: 'value', value: V} |
   {type: 'error', value: any} |
   {type: 'end', value: void};
 
-export type Emitter<T> = {
-  value(value: T): boolean;
-  event(event: Event<T>): boolean;
+export type Emitter<V> = {
+  value(value: V): boolean;
+  event(event: Event<V>): boolean;
   error(e: any): boolean;
   end(): void;
 
   // Deprecated methods
-  emit(value: T): boolean;
-  emitEvent(event: Event<T>): boolean;
+  emit(value: V): boolean;
+  emitEvent(event: Event<V>): boolean;
 };
 
 export type Subscription = {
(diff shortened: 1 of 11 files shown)


rpominov:
That's great! Thank you for doing this!

> The file kefir.js.flow needs to be copied into the dist/ directory (where Flow will look for it) on build. I'm not sure how best to set this up in Kefir's build process.

Since we use `grunt` maybe we could use some grunt-plugin that copies files? So running `$ grunt` would also copy the file. It seems like the most idiomatic approach for our current setup.

> Optional: a test could be added that checks that the kefir.js.flow file is parseable, and verifies that Flow can typecheck a test file. We should add "flow-bin" as a devDependency if we want to do that. I could take a stab at this.

Sounds like a good idea.

kefir.js.flow
```diff
@@ -0,0 +1,160 @@
+/* @flow */
+
+export type Event<T> =
+  {type: 'value', value: T} |
+  {type: 'error', value: any} |
```
rpominov: I think error type should also be parameterized. Did you see https://github.com/flowtype/flow-typed/blob/master/definitions/npm/kefir_v3.x.x/flow_%3E%3Dv0.20.x/kefir_v3.x.x.js btw?

kefir.js.flow
```diff
@@ -0,0 +1,160 @@
...
+};
+
+declare class Observable<+T> {
+  toProperty(): Property<T>;
+  toProperty<U>(getCurrent: ?() => U): Property<T|U>;
+  changes(): Observable<T>;
+
+  observe(obs: Observer<T>): Subscription;
+  observe(onValue: ?(i: T) => void, onError: ?(err: any) => void, onEnd: ?() => void): Subscription;
+  onValue(cb: (i: T) => void): Observable<T>;
```
rpominov: I never used this flow feature before, but I think `Observable<T>` can be replaced with `this` in this case.

kefir.js.flow
AgentME: I've typed the error type, and made it default to being inferred so that it doesn't need to be specified.

I hadn't seen that flow-typed declaration file before. Handy reference to compare to! Here's my notes about comparing theirs to mine. Feel free to read or skip depending on your interest level.

One difference is that theirs tries harder to keep track of properties and streams as different types (I never bothered much since Kefir has the same methods on both anyway).

It has a bunch of overloads depending on number of parameters for `Kefir.merge` and `Kefir.zip`. The Kefir.merge overloads are no longer necessary if you mark Observable's type parameter as covariant, which mine does. (This lets Flow know that it's fine to cast `Observable<Subclass>` to `Observable<Superclass>`. So if `Kefir.merge` is typed to take an array of `Observable<V>`, and you pass it an array of different types of streams, Flow automatically casts all of the streams to the union of all of their types, and the resulting merged stream has that union type.) The Kefir.zip overloads are to work around https://github.com/facebook/flow/issues/1251. (`Kefir.zip` and `Kefir.combine` have a similar type signature to `Promise.all`, but Flow doesn't support defining functions with that exact behavior. Flow's support for Promise.all's type is hard-coded.) The workaround isn't perfect, it only works for up to five parameters and is pretty verbose. (They could have done the same thing for Kefir.combine but didn't.) Might be worth doing.

One interesting issue with the Flow-typed declarations is in the *observable methods* combine and zip:

```js
    combine<U,F,R>(otherObs: Observable<U,F>, combinator?: (a: T, b: U) => R): Stream<R,E|F>;
    zip<U,F,R>(obs: Observable<U,F> | U[], combinator?: (a: T, b: U) => R): Stream<[T,U]|R,E|F>;
```

If combine is called without a combinator function, then it won't know what type the resulting stream is. If zip is called with a combinator function, then Flow will think the stream's values are either of the type returned by the combinator, or an 2-item array of both streams' types. (For example, if the combinator returned a string, you called onValue on the resulting stream, and in the callback tried to use the value as a string, then you'd get a Flow error if your code didn't also support arrays.)

My version defines separate overloads for with and without the combinator function which avoids that issue:

```js
  combine<V2,E2>(otherObs: Observable<V2,E2>): Observable<[V,V2],E|E2>;
  combine<V2,V3,E2>(otherObs: Observable<V2,E2>, combinator: (v: V, v2: V2) => V3): Observable<V3,E|E2>;
  zip<V2,E2>(otherObs: Observable<V2,E2>): Observable<[V,V2],E|E2>;
  zip<V2,V3,E2>(otherObs: Observable<V2,E2>, combinator: (v: V, v2: V2) => V3): Observable<V3,E|E2>;
```

rpominov:
GitHub have collapsed comments in the code, so I'm answering here for convenience.

> I've typed the error type, and made it default to being inferred so that it doesn't need to be specified.

Great!

> One difference is that theirs tries harder to keep track of properties and streams as different types (I never bothered much since Kefir has the same methods on both anyway).

Yea, seems reasonable.

> The Kefir.zip overloads are to work around facebook/flow#1251. (Kefir.zip and Kefir.combine have a similar type signature to Promise.all, but Flow doesn't support defining functions with that exact behavior. Flow's support for Promise.all's type is hard-coded.) The workaround isn't perfect, it only works for up to five parameters and is pretty verbose. (They could have done the same thing for Kefir.combine but didn't.) Might be worth doing.

Yea, I think it worth doing.

> My version defines separate overloads for with and without the combinator function which avoids that issue

Great!

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+  skipUntilBy(otherObs: Observable<any,any>): Observable<V,E>;
+  takeUntilBy(otherObs: Observable<any,any>): Observable<V,E>;
+  bufferBy(otherObs: Observable<any,any>, options?: {flushOnEnd?: boolean}): Observable<V[],E>;
+  bufferWhileBy(otherObs: Observable<any,any>, options?: {flushOnEnd?: boolean, flushOnChange?: boolean}): Observable<V[],E>;
+  sampledBy(otherObs: Observable<any,any>): Observable<V,E>;
+  sampledBy<V2,V3>(otherObs: Observable<V2,any>, combinator: (obsValue: V, otherObsValue: V2) => V3): Observable<V3,E>;
+  bufferWithTimeOrCount(time: number, count: number, options?: {flushOnEnd: boolean}): Observable<V,E>;
+}
+
+declare class Pool<V,E=*> extends Observable<V,E> {
```
rpominov: Wow, did't know you can set default value like this. Is there any docs about this feature?

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+
+declare var Kefir: {
+  Observable: typeof Observable;
+  Pool: typeof Pool;
+  Stream: typeof Stream;
+  Property: typeof Property;
+
+  staticLand: {
+    Observable: {
+      ap: Function; //TODO replace Function with more specific types
```
rpominov: Just in case you're struggling: `<A,B,E1,E2>(Observable<(x: A) => B, E1>, Observable<A, E2>): Observable<B, E1 | E2>`

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+      bimap: Function;
```
rpominov: `<V1,E1,V2,E2>((x: E1) => E2, (x: V1) => V2, Observable<V1,E1>): Observable<V2,E2>`

kefir.js.flow
AgentME: I'm not sure if it's in the documentation yet. I think I saw it in the changelog or release blog.

rpominov: Found it https://github.com/facebook/flow/releases/tag/v0.25.0 . Good to know.

AgentME:
I added a bunch of tests for the flow types. They're just javascript files under test/flow/ that have some flow types and lines that are expected to trigger errors. Turns out Flow supports marking lines expected to throw errors, so no special setup was needed besides adding "flow check" to the "test" script in package.json.

AgentME:
I started going through the effort of adding some overloads for Kefir.combine and Kefir.zip to make its types fully correct with up to 5 parameters, but then I realized that there's also that second passive observables array parameter to Kefir.combine, so there'd have to be an overload for every combination of number of active and passive observables up to 5 to be fully right. And none of that works right if the number and types of array items isn't known to Flow. I'm thinking it's better and more predictable for users to just have the consistently general/weaker type signature that covers all cases than spend a lot of space special-casing only some of them.

AgentME:
I've set it up so the build-kefir task calls the copy:flow task which copies kefir.js.flow into dist/ on build. I believe this pull request is ready now.

AgentME:
Optionally, we could add a note to the readme about the presence of the bundled Flow types and give a shout-out to Flow. I've been sticking this blurb at the end of some of my modules' readmes (from [redux-action-log](https://github.com/AgentME/redux-action-log)):

>## Types

>[Flow](https://flowtype.org/) type declarations for this module are included! If you are using Flow, they won't require any configuration to use.

rpominov:
> so there'd have to be an overload for every combination of number of active and passive observables up to 5

Ah, right. Ok, let's not do this then.

> Turns out Flow supports marking lines expected to throw errors, so no special setup was needed besides adding "flow check" to the "test" script in package.json.

But will build fail if line won't throw?

> Optionally, we could add a note to the readme about the presence of the bundled Flow declarations 

Good idea! Also I think a small example will be helpful. I did it like [this](https://github.com/rpominov/fun-task#flow) in fun-task.

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+  Pool: typeof Pool;
+  Stream: typeof Stream;
+  Property: typeof Property;
+
+  staticLand: {
+    Observable: {
+      ap<A,B,E1,E2>(obsF: Observable<(x: A) => B, E1>, obsV: Observable<A, E2>): Observable<B,E1|E2>;
+      bimap<V1,E1,V2,E2>(fnE: (x: E1) => E2, fnV: (x: V1) => V2, obs: Observable<V1,E1>): Observable<V2,E2>;
+      chain<V,V2,E,E2>(cb: (value: V) => Observable<V2,E2>, s: Observable<V,E>): Observable<V2,E|E2>;
+      concat<V,E>(obss: Observable<V,E>[]): Observable<V,E>;
```
rpominov: `<V1,E1,V2,E2>(obs1: Observable<V1,E1>, obs2: Observable<V2,E2>): Observable<V1|V2,E1|E2>`

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+  emitEvent(event: Event<V,E>): boolean;
+};
+
+export type Subscription = {
+  closed: boolean;
+  unsubscribe(): void;
+};
+
+export type Observer<V,E> = {
+  value(value: V): void;
```
rpominov: These properties should be optional.

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+  error(err: E): void;
+  end(): void;
+};
+
+declare class Observable<+V,+E=*> {
+  toProperty(): Property<V,E>;
+  toProperty<V2>(getCurrent: ?() => V2): Property<V|V2,E>;
```
rpominov: Does it need to be optional since first overload covers the case when argument not passed?

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+
+  map<V2>(cb: (v: V) => V2): Observable<V2,E>;
+  filter(cb?: (v: V) => any): Observable<V,E>;
+  take(n: number): Observable<V,E>;
+  takeWhile(cb?: (v: V) => boolean): Observable<V,E>;
+  last(): Observable<V,E>;
+  skip(n: number): Observable<V,E>;
+  skipWhile(cb?: (v: V) => boolean): Observable<V,E>;
+  skipDuplicates(comparator?: (a: V, b: V) => boolean): Observable<V,E>;
+  diff(fn?: (prev: V, next: V) => V, seed?: V): Observable<V,E>;
```
rpominov: Same here with overloads. I just not sure Flow will be able to choose correct overload if both take same arguments.

AgentME:
>But will build fail if line won't throw?

Tests (npm test) will fail if any of the expected flow errors are missing.

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+}
+
+declare class Stream<+V,+E=*> extends Observable<V,E> {
+}
+
+declare class Property<+V,+E=*> extends Observable<V,E> {
+}
+
+declare var Kefir: {
+  Observable: typeof Observable;
```
rpominov: Maybe we need `Class<Observable>` here? Or is it the same thing as `typeof Observable` in this case?

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+  fromPoll<V>(interval: number, f: () => V): Observable<V,*>;
+  withInterval<V>(interval: number, f: (emitter: Emitter<V>) => void): Observable<V,*>;
+  fromCallback<V>(f: (cb: (value: V) => void) => void): Observable<V,*>;
+  fromNodeCallback<V,E>(f: (cb: (err: E, value: ?V) => void) => void): Observable<V,E>;
+  fromEvents(target: Object, eventName: string, transformer?: (event: any) => any): Observable<any,*>;
+  stream<V,E>(subscribe: (emitter: Emitter<V,E>) => ?() => void): Observable<V,E>;
+
+  constant<V>(value: V): Property<V,*>;
+  constantError<E>(err: E): Property<*,E>;
+  fromPromise<V>(promise: Promise<V>): Property<V,*>;
```
rpominov: `Property<V,any>` would more correct here I think.

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+
+  fromESObservable(observable: Object): Observable<any,*>;
+  toESObservable(): Object;
+
+  combine<E>(obss: Observable<any,E>[], passiveObss?: Observable<any,E>[], combinator?: Function): Observable<any,E>;
```
rpominov: Even though we don't do overloads, we can still parametrize value, right?

kefir.js.flow
```diff
@@ -0,0 +1,158 @@
...
+  zip<V,E>(obss: Observable<V,E>[]): Observable<Array<V>,E>;
+  zip<E>(obss: Observable<any,E>[], combinator: Function): Observable<any,E>;
```
rpominov: Same here, we can do better than `Function`. And in `combine` we should do overloads with and without combinator like here.

kefir.js.flow
AgentME: Flow doesn't support the ability to properly type functions like combine or Promise.all.

kefir.js.flow
AgentME: typeof Observable works. (One of the test files happens to check this.)

kefir.js.flow
rpominov: I was thinking something like this: `combine<V,V1,E>(obss: Observable<V,E>[], passiveObss: Observable<V,E>[], combinator: (...rags: V[]) => V1): Observable<V1,E>`

rpominov: We would need 3 overloads here:

```
combine(obss, passiveObss, combinator)
combine(obss, combinator)
combine(obss)
```

AgentME: In a project that had an earlier version of these Kefir declarations bundled in, I used to do that, but it never worked in a useful way. That type signature would mean in

```js
var s1 = Kefir.constant(1);
var s2 = Kefir.constant('two');
Kefir.combine([s1,s2]).onValue(([v1, v2]) => {
  //...
});
```

The type of v1 and v2 would both be `number|string`. I would always be using :any casts when I used Kefir.combine. I'm used to needing :any casts here and there, but I don't think the combine type signature is very useful if every single use needed :any casts.

rpominov: Ok, let's use any for now and see how it goes.

But we still need 3 overloads because currently we wouldn't be able to pass a function as second argument, right?

AgentME: Oh right I forgot that it could the second parameter.

AgentME: I added 

```js
  combine<E>(obss: Observable<any,E>[], combinator?: Function): Observable<any,E>;
```

If only the first (obss) parameter is passed, I think Flow picks the first overload the passes type-checks, so it's not a big deal if multiple overloads appear to work (as long as the declaration doesn't set type parameters from a missing argument). Having the overloads both support an optional combinator parameter means they could handle the parameter missing, or a variable of type `?Function` being passed in.

AgentME:
I've added a little blurb to the readme with a short code example. The goal of the code example was to show how you could explicitly type a stream, and show that the type can be automatically inferred.

rpominov:
All looks great to me now! Let me know when it's ready to merge.

AgentME:
Yeah, I think it's good to go.

rpominov:
Thanks a lot!

AgentME:
Ha sorry, actually I've got one more thing to add. Just a small nice-to-have: make it so Flow knows `.filter(Boolean)` means the stream has no falsey values.

AgentME:
I thought it wasn't going to work, but I just realized I was testing with an old version of Flow.

AgentME:
(I'll put up a new pull request in a moment.)

rpominov:
I've already released `3.5.0` 😄 
But yeah, just open another PR.

Btw, could you check that npm package works (types are exposed correctly)?

AgentME:
The npm package works great, Flow automatically picked up the types now.

PR #218, opened 2016-08-29
flow: .filter() and .filter(Boolean) removes falsey values from stream

AgentME:
This makes it so Flow knows that calling `.filter()` or `.filter(Boolean)` on a stream returns a new one with the falsey values removed.

It's based on Flow's similar recent support for `.filter(Boolean)` on arrays. I originally wanted to get this in with my last pull request, but I was just testing with an old version of Flow and thought it wasn't going to work.


rpominov:
Thanks. It's in `3.5.1`.

Issue #173, opened 2015-12-10
.log() activation


iofjuupasli:
I think there is two use of `.log`:
1. as output. So when we `.log` stream, it should be activated to perform tasks and output results to console
2. for debugging (through logging, but anyway). It shouldn't touch anything, and shouldn't activate stream, just log values when they are pass through

Almost always log is used for logging for debug purposes. And if activation\deactivation matters, after removing logging, or adding logging for debugging somewhere on deactivated stream, it starts to behave differently than before logging.

I think, It will be very useful to have a passive `.log`

jensklose:
+1

rpominov:
Hm, I think Bacon has something similar as an undocumented method `Bacon.spy(obs)`, I don't know much about it though (since it undocumented). Would be cool if someone could investigate how it works in Bacon, and in other libs if they have similar features.

How do you think it should work? There is at lest two options:

```js
// 1. Return a new observable.
//    Basically `obs.map(console.log)`, but with output like in `obs.log()`
const newObs = obs.spy();
newObs.onValue();

// 2. Just spies on an observable, 
//    allowing us to work with same obs instance in the rest of the code
obs.spy();
obs.onValue();
```

First one probably much easier to implement.

Not sure yet if we should add this, although don't see strong reasons why not.

iofjuupasli:
The first as I understand can be implemented using `withHandler` which reemits all events.

#147 
Now `withHandler` on property returns stream. It's the only problem I see with this approach.

If it will return observable of same type, I think there is no difference for user whether it's the same object or another with same behavior

AlexGalays:
Just got bitten by this; Had no idea log() would activate a Property and make the subsequent onValue not trigger immediately.

jeron-diovis:
+1 for this issue. `log` is a debugging tool, but it easily creates new bugs itself. It's so painful.

rpominov:
Anyone want to create a PR with `spy()` method?

polytypic:
It is called [doLog](https://github.com/baconjs/bacon.js/#observable-dolog) in Bacon.

mAAdhaTTah:
`spy` landed, so closing this one.

PR #222, opened 2016-10-07
#173 Add `.spy()` and `.offSpy()` helpers.

32bitkid:
#173 `.spy()` works just like `.log()`, except does not cause side-effects/
activation. It allows you to passively watch the events generated by an
observable, without effecting its behavior.


32bitkid:
I didn't see any existing test coverage for `.log()` and `.offLog()`; but if I missed them, let me know and I'll resolve the lack of tests.

rpominov:
Thank you for the PR!

To be honest I like more the first approach suggested here https://github.com/rpominov/kefir/issues/173#issuecomment-164140556 . In Bacon `doLog` also works like that, btw. It can be implemented using `withHandler`.

And yeah, test would be nice.

mAAdhaTTah:
What about just changing `log` to not activate the stream? Or would that be BC-breaking? If `spy` existed, I would never use `log`.

rpominov:
I think current behavior of `log` is also useful. For example it's used in docs a lot.

32bitkid:
I get what you are saying, and I considered using a wrapper on top `withHandler` but that means that you can't start and stop spying on observable chains that have already been constructed.

For example:

```
// generator.js
import Kefir from 'kefir';
export default Kefir.sequentially(1000, [1,2,3,4,5]);

//squared.js
import source from './generator';
export default source.map(x => x * x);

// display.js
import result from './squared.js'
result.observe({ value(x) { render(x); }});
```

You couldn't just do something like  `require('generator').spy()` from the console at runtime, because that would create a new terminal point that has no subscribers. By making spies a first class citizen you can spy/peek on an observable you have a reference to, whether or not its in the activation chain or not.

At least, that was my rationale when thinking about it. 😄

rpominov:
Good point, @32bitkid . Let me think a bit more about it.

docs-src/descriptions/main-methods.jade
```diff
@@ -174,3 +174,30 @@ pre(title='console output')
...
+  #[b .log], however #[b .spy] will not cause the stream to activate by itself.
+  Accepts an optional #[b name] argument that will be shown in the log if provided.
+
+pre.javascript(title='example')
+  :escapehtml
+    var wait = Kefir.later(1000, 1);
+    var stream = Kefir.sequentially(250, [2, 3]);
+    stream.spy('spied'); // stream is not activated here.
+    var both = Kefir.concat(wait, stream);
+    both.log('logger');
```
rpominov: Let simplify this example? Something like:

```js
var stream = Kefir.sequentially(250, [2, 3]);
stream.spy('spied'); // stream is not activated here.
stream.observe(() => {});
```

src/dispatcher.js
```diff
@@ -43,8 +44,23 @@ extend(Dispatcher.prototype, {
     return this._items.length;
   },
 
+
+  addSpy(fn) {
+    this._spies = concat(this._spies, [fn]);
+    return this._spies.length;
+  },
+
+  removeSpy(fn) {
```
rpominov: Could you add a code comment here saying something like "we don't need complicated logic like in remove() method here, because we don't allow adding arbitrary functions as spies, our spies always functions that do only logging"?

rpominov:
Ok, let's do it this way. 

Still we need tests for `spy`, even though we don't have them for `log`. This is more complicated than `log`. And we need tests for `log` as well, tbh, I was just too lazy and didn't add them.

32bitkid:
sounds good, i'll update the pull request soon! i'll update the pull-request soon.

32bitkid:
I think that's it. let me know if there is anything else I can do to help!

rpominov:
Everything looks great, thank you!

rpominov:
Version `3.6.0` is out with this change.

PR #229, opened 2016-11-29
Flow definition fix and updates

AgentME:
* The definition of the `withInterval` method was broken. ([Flow seems to only report the error if you use the method, or as of the v0.36, pass a reference to the Kefir object](https://github.com/facebook/flow/issues/2913).)
* Updates to keep compatibility with the latest version of Flow. I figure it's intended that the `observe` method can be passed an instance of a class with value/error/end methods, but instances of classes now have their methods typed as covariant, so the Observer type needed to have its properties marked as covariant (`+` prefix does that now).
* The new version of Flow now often picks the wrong overload of `observe`, because it believes a function could be an Observer object (that just happens to be callable and has none of the value/error/end properties). [There isn't currently a way to specify that the Observer type shouldn't be callable](https://github.com/facebook/flow/issues/2914), so I made it so Flow knows that the Observer object overload can only take one parameter (Flow allows functions to be passed too many arguments by default; I disallow that here by annotating all arguments after the first as needing to be the impossible `empty` type). There's some definition checks added for observe for this.

### DIFF
--- a/kefir.js.flow
+++ b/kefir.js.flow
@@ -136,7 +136,7 @@ declare var Kefir: {
   interval<V>(interval: number, value: V): Observable<V,*>;
   sequentially<V>(interval: number, values: V[]): Observable<V,*>;
   fromPoll<V>(interval: number, f: () => V): Observable<V,*>;
-  withInterval<V>(interval: number, f: (emitter: Emitter<V>) => void): Observable<V,*>;
+  withInterval<V,E>(interval: number, f: (emitter: Emitter<V,E>) => void): Observable<V,E>;
   fromCallback<V>(f: (cb: (value: V) => void) => void): Observable<V,*>;
   fromNodeCallback<V,E>(f: (cb: (err: E, value: ?V) => void) => void): Observable<V,E>;
   fromEvents(target: Object, eventName: string, transformer?: (event: any) => any): Observable<any,*>;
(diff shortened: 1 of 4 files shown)


kefir.js.flow
```diff
@@ -22,17 +22,17 @@ export type Subscription = {
...
+  +end?: () => void;
 };
 
 declare class Observable<+V,+E=*> {
   toProperty(): Property<V,E>;
   toProperty<V2>(getCurrent: () => V2): Property<V|V2,E>;
   changes(): Observable<V,E>;
 
-  observe(obs: Observer<V,E>): Subscription;
+  observe(obs: Observer<V,E>, ...none: empty[]): Subscription;
```
rpominov: Maybe we should also change order of overloads in addition (or instead) to this? To support `obs.observe(x => { /* my on value handler */ })` case.

AgentME: I don't think the order of the overloads makes a difference; I tried it both ways while tweaking the tests.

The function-parameters overload still works if the user annotates the callback's parameter:

```js
obs.observe((x: number) => { /* my on value handler */ })
```

rpominov:
I'll merge and release a new version later today, when I have a better internet connection.

rpominov:
3.6.1 is out. 

Thank you for the PR!

PR #232, opened 2017-01-15
Add typings for setName

mAAdhaTTah:


kefir.js.flow
```diff
@@ -46,6 +46,8 @@ declare class Observable<+V,+E=*> {
   offLog(name?: string): this;
   spy(name?: string): this;
   offSpy(name?: string): this;
+  setName(name?: string): this;
```
rpominov: `name` is not optional here.

kefir.js.flow
```diff
@@ -46,6 +46,8 @@ declare class Observable<+V,+E=*> {
...
+  setName(Observable<V, E>, name?: string): this;
```
rpominov: I think this should be `setName(Observable<any, any>, name: string): this` i.e., say an `Observable` of `number` can be created from an `Observable` of another type, not necessarily `number`. And  `name` also not optional.

mAAdhaTTah:
Updated.

rpominov:
Thanks! I'll release a version with this change soon.

rpominov:
`3.7.1` is out! Sorry for the delay. And thanks again for the PR!

Issue #235, opened 2017-02-15
flatMapLatest – add next observable before removing previous


32bitkid:
In `AbstractPool` there is an option `drop` that is used by `flatMapLatest()`, when `drop` is set to `old` then the old source is unplugged from the pool before the next source is added:

```
      /* snip */
      } else if (this._drop === 'old') {
        this._removeOldest();
        this._add(obj, toObs);
      }
      /* snip */
```
https://github.com/rpominov/kefir/blob/master/src/many-sources/abstract-pool.js#L37

Is there a strong opinion against either a) adding another option, perhaps `old-soft`, to softly drop the old stream (add the next stream before removing the eldest) or b) making the default behavior for `old` add the next stream before removing the old stream.

Option (b) seems perhaps a little risky and a breaking change. But having control over this would be beneficial in some cases where you are switching between streams, but don't want them to deactivate it unless they are unused in the next streams activation chain.

Example:

```
const channelA = Kefir.stream((emitter) => {
  console.log('connect a');
  let count = 0, id = setInterval(() => emitter.value(count++), 250);
  return () => { console.log('disconnect a'); clearInterval(id); };
});

const channelB = Kefir.stream((emitter) => {
  console.log('connect b');
  let count = 0, id = setInterval(() => emitter.value(count++), 250);  
  return () => { console.log('disconnect b'); clearInterval(id); };
});

const data = {
  a: channelA,
  b: Kefir.combine([channelA, channelB]),
  c: channelB,
};

Kefir.sequentially(1000, ['a', 'b', 'c', undefined])
  .flatMapLatest(p => p ? data[p] : Kefir.never())
  .log('result');
```

Console:
```
> connect a
> result <value> 0
> result <value> 1
> result <value> 2
> disconnect a
> connect a
> connect b
> result <value> [0, 0]
> result <value> [1, 0]
> result <value> [1, 1]
> result <value> [2, 1]
> result <value> [2, 2]
> disconnect a
> disconnect b
> connect b
> result <value> 0
> result <value> 1
> result <value> 2
> disconnect b
> result <end>
```

Notice how both `a` and `b` disconnect and immediately reconnect on the swap. What, I think, I would like in some situations:
```
> connect a
> result <value> 0
> result <value> 1
> result <value> 2
> connect b
> result <value> [3, 0]
> result <value> [3, 1]
> result <value> [4, 1]
> result <value> [4, 2]
> disconnect a
> connect b
> result <value> 3
> result <value> 4
[...]

PR #236, opened 2017-02-16
#235 flatMapLatest adding overlapping option

32bitkid:
passing an options object of {overlapping:true} will cause the next stream to be added before any old streams are removed.

### DIFF
--- a/src/many-sources/abstract-pool.js
+++ b/src/many-sources/abstract-pool.js
@@ -5,12 +5,13 @@ import {concat, forEach, findByPred, find, remove, cloneArray} from '../utils/co
 
 const id = (x => x);
 
-function AbstractPool({queueLim = 0, concurLim = -1, drop = 'new'} = {}) {
+function AbstractPool({queueLim = 0, concurLim = -1, drop = 'new', overlapping = false} = {}) {
   Stream.call(this);
 
   this._queueLim = queueLim < 0 ? -1 : queueLim;
   this._concurLim = concurLim < 0 ? -1 : concurLim;
   this._drop = drop;
+  this._overlapping = overlapping;
   this._queue = [];
   this._curSources = [];
   this._$handleSubAny = (event) => this._handleSubAny(event);
@@ -34,8 +35,15 @@ inherit(AbstractPool, Stream, {
       if (this._queueLim === -1 || this._queue.length < this._queueLim) {
         this._addToQueue(toObs(obj));
       } else if (this._drop === 'old') {
-        this._removeOldest();
-        this._add(obj, toObs);
+        if (this._overlapping) {
+          this._addToCur(toObs(obj));
+          while (this._curSources.length > this._concurLim) {
+            this._removeOldest();
+          }
+        } else {
+          this._removeOldest();
+          this._add(obj, toObs);
+        }
       }
     }
   },
--- a/docs-src/descriptions/multiple-sources.jade
+++ b/docs-src/descriptions/multiple-sources.jade
@@ -403,7 +403,7 @@ div
 
 
 
-+descr-method('flat-map-latest', 'flatMapLatest', 'obs.flatMapLatest([fn])').
++descr-method('flat-map-latest', 'flatMapLatest', 'obs.flatMapLatest([fn], [options])').
   Like #[b flatMap], but repeats events only from the latest added observable
   i.e., switching from one observable to another.
 
@@ -433,6 +433,14 @@ pre(title='events in time').
   spawned 3:                                 ---3---3---3---3X
 
   result:      -------------1---1-----2---2-----3---3---3---3X
+
+p.
+  By default, #[b flatMapLatest] will remove the previous observable #[i before]
+  adding the next. This may cause an obversable to deactivate briefly when switching
+  between observable, even if the observable is in the activation chain of both the
+  previous and next observable. If you want the next observable to be added #[i before]
+  removing the previous, you can pass #[tt {overlapping: true}] as #[b options].
+
 div
 
 
--- a/kefir.js.flow
+++ b/kefir.js.flow
@@ -89,7 +89,8 @@ declare class Observable<+V,+E=*> {
   concat<V2,E2>(otherObs: Observable<V2,E2>): Observable<V|V2,E|E2>;
 
   flatMap<V2,E2>(transform: (value: V) => Observable<V2,E2>): Observable<V2,E|E2>;
-  flatMapLatest<V2,E2>(transform: (value: V) => Observable<V2,E2>): Observable<V2,E|E2>;
+  flatMapLatest<V2,E2>(options?: {overlapping?: boolean}): Observable<V2,E|E2>;
+  flatMapLatest<V2,E2>(transform: (value: V) => Observable<V2,E2>, options?: {overlapping?: boolean}): Observable<V2,E|E2>;
   flatMapFirst<V2,E2>(transform: (value: V) => Observable<V2,E2>): Observable<V2,E|E2>;
   flatMapConcat<V2,E2>(transform: (value: V) => Observable<V2,E2>): Observable<V2,E|E2>;
   flatMapConcurLimit<V2,E2>(transform: (value: V) => Observable<V2,E2>, limit: number): Observable<V2,E|E2>;
--- a/src/index.js
+++ b/src/index.js
@@ -332,8 +332,10 @@ import FlatMap from './many-sources/flat-map';
 Observable.prototype.flatMap = function(fn) {
   return new FlatMap(this, fn).setName(this, 'flatMap');
 };
-Observable.prototype.flatMapLatest = function(fn) {
-  return new FlatMap(this, fn, {concurLim: 1, drop: 'old'}).setName(this, 'flatMapLatest');
+
+import flatMapLatest from './many-sources/flat-map-latest';
+Observable.prototype.flatMapLatest = function(...args) {
+  return flatMapLatest(this, ...args);
 };
 Observable.prototype.flatMapFirst = function(fn) {
   return new FlatMap(this, fn, {concurLim: 1}).setName(this, 'flatMapFirst');
--- a/src/many-sources/flat-map-latest.js
+++ b/src/many-sources/flat-map-latest.js
@@ -1,0 +1,11 @@
+import FlatMap from './flat-map';
+
+export default function flatMapLatest(obs, fn, options) {
+  if (typeof fn !== 'function') {
+    options = fn;
+    fn = undefined;
+  }
+  options = options === undefined ? {} : options;
+  const { overlapping = false } = options;
+  return new FlatMap(obs, fn, {concurLim: 1, drop: 'old', overlapping }).setName(obs, 'flatMapLatest');
+}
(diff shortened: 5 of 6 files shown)


src/many-sources/abstract-pool.js
```diff
@@ -34,8 +35,15 @@ inherit(AbstractPool, Stream, {
       if (this._queueLim === -1 || this._queue.length < this._queueLim) {
         this._addToQueue(toObs(obj));
       } else if (this._drop === 'old') {
-        this._removeOldest();
-        this._add(obj, toObs);
+        if (this._overlapping) {
+          this._addToCur(toObs(obj));
+          while (this._curSources.length > this._concurLim) {
+            this._removeOldest();
+          }
```
32bitkid: this loop seemed a little weird when I was working on it, however, I think to have correct parity with the other side of the condition it needs to be here. Given that `this._curSources.length - this._concurLim > 1`, the other side, implicitly loops due to recursion. Since we want to add _first_, using `_addToCur`, then we need to make sure that we aren't over the `_concurLim` and blindly removing one might be not enough. Although, I'm not sure how an `AbstractPool` could get into that state.

rpominov:
Thank you for the PR! I'll take a look tomorrow. And sorry for the delays, quite busy at work.

src/many-sources/abstract-pool.js
rpominov: I think we should just remove one. Yes, there is a recursion on the other side, but it should always do only one iteration because:
1. We always expect this condition to be true: `this._curSources.length <= this._concurLim`.
2. When we get to this branch of the code `this._curSources.length === this._concurLim` should be true.
3. We remove one from `this._curSources`, so now `this._curSources.length === this._concurLim - 1`.
4. At the following `this._add()` call, we fall into branch at line 33. And that is the end.

But more importantly, with the condition you added we might actually remove none of the oldest. After `this._addToCur(toObs(obj))` we not necessarily have an observable added to `_curSources` because if `toObs(obj)` already anded or ends immediately after we subscribe to it, we don't add it to the `_curSources` (take a look at `_addToCur`). Btw, we need a test case for this.

rpominov: Also maybe it make sense to remove recursion on other side too.

```diff
  this._removeOldest();
- this._add(obj, toObs);
+ this._addToCur(toObs(obj));
```

This way we get the correct parity.

32bitkid: I'm not thats quite equivalent, I need to think about it more, and there may not be a public api to aggravate the condition, but I think that when `{ concurLim: 1, queueLim: 1, drop: 'old' }`, and given that both `_curSources` and `_queue` are full and I add another one, I would want to:

1. drop the eldest item out of `_curSources`
2. promote the eldest item from `_queue`
3. add_ the new_ stream to the `_queue`

But, I'll need to think about it more. However, I think the same concern is valid for the overlapping side of the branch. If there _is_ a queue, then one should add to `_queue` and let `_removeOldest` pull it out of the queue, but that doesn't really resolve the actual use-case for using overlapping to begin with. I think I need to step back and think about it harder.

src/many-sources/abstract-pool.js
```diff
@@ -37,12 +37,10 @@ inherit(AbstractPool, Stream, {
...
         if (this._overlapping) {
           this._addToCur(toObs(obj));
-          while (this._curSources.length > this._concurLim) {
-            this._removeOldest();
-          }
+          this._removeOldest();
         } else {
           this._removeOldest();
-          this._add(obj, toObs);
+          this._addToCur(toObs(obj));
```
32bitkid: while cleaner; this doesn't _feel_ right. see my previous comment, I need to think more about this/this is idea is more subtle than I initially projected.

src/many-sources/abstract-pool.js
rpominov: Ah, right! I didn't realize that `_removeOldest` pulls from the `queue`. Replacing `this._add -> this._addToCur` isn't equivalent then indeed. I'll think about it too.

rpominov: I haven't test it, but maybe we could do something like this?

```js
_add(obj, toObs /* Function | falsey */, allowOverflow) {
  toObs = toObs || id;
  if (this._concurLim === -1 || this._curSources.length < this._concurLim) {
    this._addToCur(toObs(obj));
  } else {
    if (this._queueLim === -1 || this._queue.length < this._queueLim || allowOverflow) {
      this._addToQueue(toObs(obj));
    } else if (this._drop === 'old') {
      if (this._overlapping) {
        this._add(obj, toObs, true);
        this._removeOldest();
      } else {
        this._removeOldest();
        this._add(obj, toObs);
      }
    }
  }
}
```

mAAdhaTTah:
This would be a nice-to-have. Anything blocking?

mAAdhaTTah:
@32bitkid Would love to land this, if possible. There's been some changes to our test suite in particular (it's in JS now woo!), so the `.coffee` changes would need to be ported over, and there are a couple other conflicts. Is this something you're still interested in pushing?

rpominov:
@mAAdhaTTah There are also some unresolved concerns in folded comments.

32bitkid:
@mAAdhaTTah This is a feature that I'd really like to get in as well, but have been swamped with work lately. As @rpominov mentioned, there are some unresolved issues that still need to be addressed/reworked. I'll try to spend some time this weekend on this.

bpinto:
This PR is now 4 years old, I worked on a patch on the current codebase, and I wonder if I should open a PR with it. What do you think?

```
diff --git a/src/index.js b/src/index.js
index daf4001..5eaf803 100644
--- a/src/index.js
+++ b/src/index.js
@@ -331,8 +331,9 @@ import FlatMap from './many-sources/flat-map'
 Observable.prototype.flatMap = function(fn) {
   return new FlatMap(this, fn).setName(this, 'flatMap')
 }
-Observable.prototype.flatMapLatest = function(fn) {
-  return new FlatMap(this, fn, {concurLim: 1, drop: 'old'}).setName(this, 'flatMapLatest')
+import flatMapLatest from './many-sources/flat-map-latest'
+Observable.prototype.flatMapLatest = function(...args) {
+  return flatMapLatest(this, ...args);
 }
 Observable.prototype.flatMapFirst = function(fn) {
   return new FlatMap(this, fn, {concurLim: 1}).setName(this, 'flatMapFirst')
diff --git a/src/many-sources/abstract-pool.js b/src/many-sources/abstract-pool.js
index 683314c..19865db 100644
--- a/src/many-sources/abstract-pool.js
+++ b/src/many-sources/abstract-pool.js
@@ -5,12 +5,13 @@ import {concat, forEach, findByPred, find, remove, cloneArray} from '../utils/co
 
 const id = x => x
 
-function AbstractPool({queueLim = 0, concurLim = -1, drop = 'new'} = {}) {
+function AbstractPool({queueLim = 0, concurLim = -1, drop = 'new', overlapping = false} = {}) {
   Stream.call(this)
 
   this._queueLim = queueLim < 0 ? -1 : queueLim
   this._concurLim = concurLim < 0 ? -1 : concurLim
   this._drop = drop
+  this._overlapping = overlapping
   this._queue = []
   this._curSources = []
   this._$handleSubAny = event => this._handleSubAny(event)
@@ -33,8 +34,13 @@ inherit(AbstractPool, Stream, {
       if (this._queueLim === -1 || this._queue.length < this._queueLim) {
         this._addToQueue(toObs(obj))
       } else if (this._drop === 'old') {
-        this._removeOldest()
-        this._add(obj, toObs)
+        if (this._overlapping) {
+          this._addToCur(obj, toObs)
+          this._removeOldest()
+        } else {
+          this._removeOldest()
+          this._add(obj, toObs)
+        }
       }
     }
   },
diff --git a/src/many-sources/flat-map-latest.js b/src/many-sources/flat-map-latest.js
new file mode 100644
index 0000000..0439943
--- /dev/null
+++ b/src/many-sources/flat-map-latest.js
@@ -0,0 +1,21 @@
+import FlatMap from './flat-map'
+
+function flatMapLatest(source, fn, options) {
+  //console.log({ source })
+  //console.log({ fn })
+  //console.log({ options })
+  if (typeof fn !== 'function') {
+    options = fn;
+    fn = undefined;
+  }
+
+  //console.log({ source })
+  //console.log({ fn })
+  //console.log({ options })
+
+  options = options === undefined ? {} : options;
+  const { overlapping = false } = options;
+  return new FlatMap(source, fn, {concurLim: 1, drop: 'old', overlapping }).setName(source, 'flatMapLatest');
+}
+
+export default flatMapLatest
diff --git a/test/specs/flat-map-latest.js b/test/specs/flat-map-latest.js
index 3893b39..9c2c0e1 100644
--- a/test/specs/flat-map-latest.js
+++ b/test/specs/flat-map-latest.js
@@ -1,3 +1,4 @@
+const sinon = require('sinon')
 const {stream, prop, send, value, error, end, activate, deactivate, Kefir, expect} = require('../test-helpers')
 
 describe('flatMapLatest', () => {
@@ -82,6 +83,59 @@ describe('flatMapLatest', () => {
         })
       ).to.emit([value(3), value(4), value(5)], () => send(a, [value(1), value(2), value(3), value(4), value(5)]))
     })
+
+    describe('non-overlapping', () => {
+      it('should remove the previous stream before adding the next', () => {
+        onDeactivate = sinon.spy()
+        a = Kefir.stream(() => onDeactivate)
+        b = stream()
+        map = b.flatMapLatest()
+        activate(map)
+        send(b, [value(a)])
+        send(b, [value(a)])
+        deactivate(map)
+        expect(onDeactivate.callCount).to.equal(2)
+      })
+    })
+
+    describe('overlapping', () => {
+      it('should add the next stream before removing the previous', () => {
+        onDeactivate = sinon.spy()
+        a = stream()
+        b = Kefir.stream(() => onDeactivate)
+        map = a.flatMapLatest({ overlapping: true })
+        activate(map)
+        send(a, [value(b)])
+        send(a, [value(b)])
+        deactivate(map)
+        expect(onDeactivate.callCount).to.equal(1)
+      })
+
+      it('should accept optional map fn', () => {
+        onDeactivate = sinon.spy()
+        a = stream()
+        b = Kefir.stream(() => onDeactivate)
+        map = a.flatMapLatest(x => x.obs, { overlapping: true })
+        activate(map)
+        send(a, [value({ obs: b })])
+        send(a, [value({ obs: b })])
+        deactivate(map)
+        expect(onDeactivate.callCount).to.equal(1)
+      })
+
+      it('should work nicely with Kefir.constant and Kefir.never', () => {
+        const a = stream()
+        expect(
+          a.flatMapLatest(x => {
+            if (x > 2) {
+              return Kefir.constant(x)
+            } else {
+              return Kefir.never()
+            }
+          }, { overlapping: true })
+        ).to.emit([value(3), value(4), value(5)], () => send(a, [value(1), value(2), value(3), value(4), value(5)]))
+      })
+    })
   })
 
   describe('property', () => {
```

mAAdhaTTah:
@bpinto Can't hurt to open a PR!

bpinto:
I'll go through the comments on this PR again and see what I can do, I believe I have a better understanding of the feature than I had when I wrote the patch. I'll open a PR then.

PR #242, opened 2017-05-21
A few improvements for Flow type definitions

hallettj:
The type definitions for Kefir are a joy to read! I learned some interesting new things - in particular the option to specify a default type for a type parameter (e.g. `declare class Observable<+V,+E=*>`), and the `$NonMaybeType<V>` utility type are features that I'm sure will come in handy.

I had a few ideas for improvements. Consider this pull request to be my suggestions.

## `empty` type parameter for vacuous observables

In cases where an observable is guaranteed to not produce either values or errors I changed the relevant type parameter from `*` to `empty`. I think that `empty` is the correct type, because it is effectively an assertion that something will not happen. And `empty` is an identity under the union operation, so certain operations will work out nicely.

For example I changed the type for `constant` to:

```js
constant<V>(value: V): Property<V,empty>;
```

That means that the type checker will allow `onError` callbacks with any type for the error value argument, which is safe because those callbacks should never be invoked. And the choice of argument type in those callbacks will not affect the type that Flow infers for error values later in the program.

If a constant property is combined with an observable that with an error type parameter of some type `E`, Flow will infer the error type of the resulting observable to be `empty | E`, which is equivalent to just `E`, which is what we want.

## more specific types for ESObservable interop

I'm planning on writing a library that consumes reactive streams, and I thought it would be nice to use the ESObservable interface so that I do not have to tie my library to one specific FRP library. I think that having a specific type for `toESObservable` will be handy.

## fix and verify types for overloaded methods and functions

I did some testing and found that overloaded signatures were not behaving as intended. I added a lot of tests, and made a bunch of changes to overloaded functions. One trick that fixed many cases was to put a dummy argument to the end of the argument list in some signatures to help Flow to select the correct overload branch. I used `$?: empty` for that purpose. For example:

```js
diff($?: empty): Observable<[V, V],E>;
diff<V2>(fn: (prev: V, next: V) => V2, $?: empty): Observable<V2,E>;
diff<V2,V3>(fn: (prev: V | V3, next: V) => V2, seed: V3): Observable<V2,E>;
diff<V3>(fn: null | void, seed: V3): Observable<[V | V3, V],E>;
```

The problem comes up when Flow encounters an invocation such as `obs.diff((x, y) => x - y)`. If the `$?: empty` argument were not present in the first signature, then the first two signatures would both be valid interpretations of that invocation. (This is because Flow permits an invocation to provide more arguments than are listed in the function signature.) Flow tends to pick the first signature; so the type of the resulting observable ends up as `Observable<[number, number], E>` when the type should be `Observable<number, E>` (assuming that the type of the input observable is `Observable<number, E>`.

Using the optional argument `$?: empty` removes the ambiguity. There is no possible argument that matches the type `empty`. So `$?: empty` effectively forces a signature to be applicable only in cases where the method is invoked with the exact number of arguments that are listed in the signature (not including the `$?: empty` argument.) I chose `$` for the parameter name is because it reminds me of the "end of input" token in a parser.

I did notice one use of the same trick in the existing signatures:

```js
observe(obs: Observer<V,E>, ...none: empty[]): Subscription;
```

I used `$?: empty` instead of `...none: empty[]` just because the former is a bit shorter.

It might be possible to remove the `$?: empty` trick in the future. I saw this in the changelog for Flow v0.46.0, which seems to imply that there might be changes to the way Flow interprets argument list lengths in an upcoming release:

> Starting in v0.47.0 we're going to complain when you call a function with more arguments than it expects. You can try it out in v0.46.0 with the .flowconfig option experimental.strict_call_arity=true. For more [check out this blog post](https://flow.org/blog/2017/05/07/Strict-Function-Call-Arity/)

I also expanded the definition of `combine` to use the newish `$ObjMap` utility type. The new type signature is enormous - but `$ObjMap` makes it possible to specify type rules very precisely when the inputs are given as objects with observables as property values.

### DIFF
--- a/kefir.js.flow
+++ b/kefir.js.flow
@@ -44,12 +44,12 @@ type ESObservable<+V,+E=*> = {
 };
 
 declare class Observable<+V,+E=*> {
-  toProperty(): Property<V,E>;
+  toProperty($?: empty): Property<V,E>;
   toProperty<V2>(getCurrent: () => V2): Property<V|V2,E>;
   changes(): Observable<V,E>;
 
-  observe(obs: Observer<V,E>, ...none: empty[]): Subscription;
-  observe(onValue: ?(v: V) => void, onError: ?(err: E) => void, onEnd: ?() => void): Subscription;
+  observe(obs: Observer<V,E>, $?: empty): Subscription;
+  observe(onValue?: (v: V) => void, onError?: (err: E) => void, onEnd?: () => void): Subscription;
   onValue(cb: (v: V) => void): this;
   offValue(cb: (v: V) => void): this;
   onError(cb: (err: E) => void): this;
@@ -76,12 +76,18 @@ declare class Observable<+V,+E=*> {
   skip(n: number): Observable<V,E>;
   skipWhile(cb?: (v: V) => boolean): Observable<V,E>;
   skipDuplicates(comparator?: (a: V, b: V) => boolean): Observable<V,E>;
-  diff(fn?: (prev: V, next: V) => V): Observable<V,E>;
-  diff<V2>(fn: (prev: V2, next: V) => V, seed: V2): Observable<V2,E>;
-  scan(cb: (prev: V, next: V) => V): Observable<V,E>;
+
+  diff($?: empty): Observable<[V, V],E>;
+  diff<V2>(fn: (prev: V, next: V) => V2, $?: empty): Observable<V2,E>;
+  diff<V2,V3>(fn: (prev: V | V3, next: V) => V2, seed: V3): Observable<V2,E>;
+  diff<V3>(fn: null | void, seed: V3): Observable<[V | V3, V],E>;
+
+  scan(cb: (prev: V, next: V) => V, $?: empty): Observable<V,E>;
   scan<V2>(cb: (prev: V2, next: V) => V2, seed: V2): Observable<V2,E>;
-  flatten(): Observable<any,E>;
+
+  flatten($?: empty): Observable<any,E>;
   flatten<V2>(transformer: (value: V) => V2[]): Observable<V2,E>;
+
   delay(n: number): Observable<V,E>;
   throttle(n: number, options?: {leading?: boolean, trailing?: boolean}): Observable<V,E>;
   debounce(n: number, options?: {immediate?: boolean}): Observable<V,E>;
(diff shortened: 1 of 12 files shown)


hallettj:
Pinging the folks who wrote and contributed to these definitions previously: @AgentME , @32bitkid , @mAAdhaTTah

mAAdhaTTah:
I'm still reading through this, and the changes themselves, but I'm not sure I agree with the types here:

```
diff<V2,V3>(fn: (prev: V | V3, next: V) => V2, seed: V3): Observable<V2,E>;
```

I'm only picking on this one since it was in your (excellent) writeup. Without having looked at the tests, I'm not sure how this can work. The return value of the `fn` becomes both the output of the Observable as well as the `prev` value for the next function call. It can't return a type of `V2` and then only accept `V | V3` as the `prev`. Additionally, the `seed` type gets output by the Observable as well, so `fn` would have to return `V2 | V3` and the method would return `Observable<V2 | V3, E>`.

I think this can be simplified by using a single type for all of that. `seed`, `prev`, the return value of `fn` _and_ the returned Observable's generic value should all be the same type imo.

rpominov:
This looks promising. I'll be busy next few days, but will try to give this PR a proper look soonish.

> The type definitions for Kefir are a joy to read!

Thanks, most of the credit goes to @AgentME , he did a great job!

hallettj:
@mAAdhaTTah I think that type is correct - `diff` does not use the output from `fn` as the input to any later `fn` invocations. The only values that are input to `fn` are values from the original observable and the seed value. For example:

```js
> Kefir.sequentially(0, [1, 2, 3, 4]).diff((prev, next) => { console.log(`fn(${prev}, ${next})`); return true }, "foo").onValue(v => console.log('value:', v))

fn(foo, 1)
value: true
fn(1, 2)
value: true
fn(2, 3)
value: true
fn(3, 4)
value: true
```

mAAdhaTTah:
@hallettj derp I confused `diff` and `scan`. nvm!

AgentME:
Looks really good! The use of `empty` is good; I think Flow didn't have it back when I first made the type definitions so I just sprinkled `*` around. I later noticed that I hadn't set up any of the overloads for `Kefir.*` functions correctly, but I had never gotten around to fixing that up.

`$ObjMap` used for `combine` seems great. You should also try using `$TupleMap` for the array-overloads of `combine`. I think it hasn't been documented yet, but it's basically just `$ObjMap` for arrays. Flow's internal typing of `Promise.all` is a [good example of its use](https://github.com/facebook/flow/blob/f35d3a96ad1fb4a985bf522ddc63ac1f66360904/lib/core.js#L662).

hallettj:
@AgentME Thanks! And thanks for the tip on `$TupleMap`. It looks like that will work nicely for several cases: `combinator` with an array of observables and no passive observables or combinator, `zip` with no combinator, `merge`, and `concat`. I would like to find a way to make `$TupleMap` work in the presence of  combinators and passive observables as well, but there are a couple of details that I don't know how to address at this time.

We want to be able to map values types from an array of input observables to positional arguments to a combinator. For `combine` that case should look something like this:

```js
// array of observables, combinator
combine: (<E, Obss: Observable<any,E>[]>(
  obss: Obss,
  combinator: Function<$observedValues<Obss>, V>
) => Observable<V,E>)

type $observedValues<Obss> = $TupleMap<Obss, <V,E>(obs: Observable<V,E>) => V>
```

Unfortunately I do not know of an existing utility type that behaves like I'm trying to make `Function` behave in that example. I think that expressing the combinator type accurately would require a utility type with two type parameters (e.g. `Function<Args, Return>`) where the first parameter is a tuple type that expresses types of positional function arguments.

The issue with the passive observable case is that I do not know of any way to express a concatenation of two tuples as a type. We want this:

```js
// array of observables, array of passive observables
combine: (<E, E2, Obss: Observable<any,E>[], PassiveObss: Observable<any,E2>[]>(
  obss: Obss,
  passiveObss: PassiveObss,
  $?: empty
) => Observable<$observedValues<Obss> ++ $observedValues<PassiveObss>, E|E2>)

type $observedValues<Obss> = $TupleMap<Obss, <V,E>(obs: Observable<V,E>) => V>
```

But again, I don't know of an existing Flow feature that does what I want `++` to do in that example.

kefir.js.flow
```diff
@@ -1,5 +1,7 @@
 /* @flow */
 
+import type EventEmitter from "events";
```
rpominov: I wonder how this import works. We import from Flow build-ins?

hallettj:
That's right, `"events"` is described in Flow's built-in types.

https://github.com/facebook/flow/blob/master/lib/node.js

rpominov:
Ah, there is a [module declaration](https://github.com/facebook/flow/blob/master/lib/node.js#L668-L675) that makes the import work. OK, this makes sense.

rpominov:
Merged and new version released. Thanks again for the PR!

PR #244, opened 2017-06-21
Swtich === to <=

mAAdhaTTah:
If the application is structured such that it's possible to set up
a cyclical dependency, and a `take` stream emits a value that
synchronously results in another value being dispatched down
the `take` stream, the `this._n--` can get called twice, pushing it
down to -1 and thus never triggering `end`. This ensures the
stream will end after the value is emitted, even if another value
was emitted in the meantime.

### DIFF
--- a/src/one-source/take-errors.js
+++ b/src/one-source/take-errors.js
@@ -11,7 +11,7 @@ const mixin = {
   _handleError(x) {
     this._n--
     this._emitError(x)
-    if (this._n === 0) {
+    if (this._n <= 0) {
       this._emitEnd()
     }
   },
--- a/src/one-source/take-errors.js
+++ b/src/one-source/take-errors.js
@@ -9,9 +9,12 @@ const mixin = {
   },
 
   _handleError(x) {
+    if (this._n === 0) {
+      return
+    }
     this._n--
     this._emitError(x)
-    if (this._n <= 0) {
+    if (this._n === 0) {
       this._emitEnd()
     }
   },
--- a/test/specs/take.coffee
+++ b/test/specs/take.coffee
@@ -1,4 +1,4 @@
-{stream, prop, send, Kefir} = require('../test-helpers.coffee')
+{stream, prop, send, Kefir, pool} = require('../test-helpers.coffee')
 
 
 
@@ -34,6 +34,13 @@ describe 'take', ->
       a = stream()
       expect(a.take(1)).errorsToFlow(a)
 
+    it 'should emit once on circular dependency', ->
+      a = pool()
+      b = a.take(1).map((x) -> x + 1)
+      a.plug(b)
+
+      expect(b).toEmit [2, '<end>'], ->
+        send(a, [1, 2, 3, 4, 5])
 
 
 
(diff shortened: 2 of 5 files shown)


mAAdhaTTah:
I think the commit message is clear, and I don't think this should
be a controversial change, but let me know if you have any questions.
I'm running into this because I have cyclical stream dependencies in
my application.

rpominov:
Sorry that you have to deal with this kind of stuff. There are probably more subtle bugs in cases when event handler emits another event, especially to the same stream. Kefir doesn't support well such cases, in general you should try to avoid them.

As for this particular bug, I'm not sure this fix is good enough. Doesn't this result in more that `N` values being emitted even though we manage to `end` after that?

rpominov:
Also we would really need a test case for this, to be safe from regressions in future.

mAAdhaTTah:
Yeah, currently, the way the architecture is structured, it's not really possible to avoid. It's definitely something I'm thinking about, but I'm not sure there's a good way around it. The issue is I'm plugging into Redux, and want my streams to be able to emit actions into Redux and respond to all actions into it, which necessitates making the dependencies circular.

> As for this particular bug, I'm not sure this fix is good enough. Doesn't this result in more that N values being emitted even though we manage to end after that?

Yes, it can, but in the current instance, if it gets multiple events, it never ends. With this, at least after getting extra events, it will eventually end once the stack is unwound

> Also we would really need a test case for this, to be safe from regressions in future.

I'll spend some time on this tonight. I resolved it in my application by calling `delay(0)` before `take(1)` which is less than ideal but resolves the issue.

mAAdhaTTah:
I'm honestly not sure there's a way to prevent n+1 values being emitted, unless we do the check before we emit the value. Would that work, e.g. `if(this._n >= 0) this._emitValue(x)`?

rpominov:
Maybe something like this will work?

```js
const mixin = {
  // ...
  _handleValue(x) {
    if (this._n === 0) {
      return
    }
    this._n--
    this._emitValue(x)
    if (this._n <= 0) {
      this._emitEnd()
    }
  },
}
```

mAAdhaTTah:
Cool, I'll try that when I write the tests.

mAAdhaTTah:
Ok, updated this w/ tests.

src/one-source/take-errors.js
```diff
@@ -11,7 +11,7 @@ const mixin = {
   _handleError(x) {
     this._n--
     this._emitError(x)
-    if (this._n === 0) {
+    if (this._n <= 0) {
```
rpominov:
```
if (this._n === 0) {
  return
}
```
... here as well

test/specs/take.coffee
```diff
@@ -34,6 +34,13 @@ describe 'take', ->
       a = stream()
       expect(a.take(1)).errorsToFlow(a)
 
+    it 'should emit once on circular dependency', ->
```
rpominov: Please add a test for `takeErrors` as well.

mAAdhaTTah:
Better late than never. Just updated!

rpominov:
It looks great now, thanks! And sorry for the delay. Releasing a new version with this soon.

PR #251, opened 2017-10-06
#250 Add es module build

### LINKED ISSUE #250: Setup esm build for tree shaking benefits
Being as this project is already using rollup, with a little extra configuration we could output an esm (es modules) build to so that consumers of the library using webpack or rollup could benefit from tree shaking any functionality they may not use. 

https://github.com/rollup/rollup/wiki/pkg.module

Even if there's not a whole lot of benefit from tree shaking, it's still a worthy cause. 

I may do a PR myself if you're open for that.

zephraph:
Fixes #250.

This outputs and es modules build so people using rollup or webpack 2+ can benefit from tree shaking. 

I will note that it's complaining about the current export setup. 

image

Straight from the rollup docs:

> It is bad practice to mix default and named exports in the same module, though it is allowed by the specification.

For module consumers to be able to treeshake effectively they'll need to consume the named exports. I don't think leaving this in is a big deal... we could configure the exports setting like they recommend. I'll leave that up to you though.


rpominov:
Looks great!

> we could configure the exports setting like they recommend

Yes, please do. 

A bit of a history why it ended up this way: when I first published Kefir, I wanted `const {Kefir} = require('kefir')` to work for some reason. And when we migrated to rollup we preserved this. 

But I don't think any more that this is important, and it breaks tree shaking. Let's get rid of it. Some people will have to update their code maybe, but it should be easy for them.

rpominov:
Although wait :)

We probably want `import Kefir from 'kefir'` to still work, and that what rollup is complaining about.

We also allow to `import {Kefir} from 'kefir'` AFAIK, which I think is ok to remove.

mAAdhaTTah:
Losing `import Kefir from 'kefir';` would be a painful breaking change, so yeah, please leave that in.

Not that I'm opposed to the PR, but I also don't think tree shaking is going to have a large impact on the size of a bundle using Kefir, because most of the weight is in prototype methods, which can't be shaken out.

rpominov:
@zephraph , could you please look if not supporting `import {Kefir} from 'kefir'` will improve anything it terms of tree shaking? And if so, remove support of it.

Otherwise I think this PR is good to go!

zephraph:
To get any benefit from tree shaking the esm build can't use `import Kefir from 'kefir'`. That doesn't mean that the current build needs to change. If you're just using babel or require statements it should work as it always has. 

I don't think we should remove the default export though, maybe just set the `exports: named` flag as rollup suggests. 

Though, I do question whether this should be released as a major following semantic versioning. People on webpack or rollup may suddenly experience a breaking change if this rolls out. There's nothing we could due to preserve the `import Kefir from 'kefir'` for those users unless there was only a default export... which I assume would break others. 

Lastly, I was trying this out last night in my build and for whatever reason babel was puking... maybe an issue with my setup. Anyway, I can't verify yet the size changes. I'll report back if I get it working, but if anyone else could try it out I'd appreciate it!

rpominov:
> I don't think we should remove the default export though, maybe just set the exports: named flag as rollup suggests.

Agree.

> Though, I think this may need to be released as a major following semantic versioning. People on webpack or rollup may suddenly experience a breaking change if this rolls out.

Why? Because webpack or rollup will pick up our new ESM build? But seems like it has both named and default exports:

```js
export { dissableDeprecationWarnings, Kefir, Observable, Stream, Property, never, later, interval,
 sequentially, fromPoll, withInterval, fromCallback, fromNodeCallback, fromEvents, stream, constant,
 constantError, fromPromise, fromESObservable, combine, zip, merge, concat$1 as concat, Pool,
 pool, repeat, staticLand };export default Kefir;
```

> Lastly, I was trying this out last night in my build and for whatever reason babel was puking... maybe an issue with my setup.

I've tried doing `npm run build` with fresh `node_modules`, works fine for me.

zephraph:
> seems like it has both named and default exports

You're absolutely right, it does have both. The docs from rollup say that...

>The wrinkle is that if you use named exports but also have a default export, a user would have to do something like this to use the default export:

```
var yourMethod = require( 'your-lib' ).yourMethod;
var yourLib = require( 'your-lib' )['default'];
```

Maybe this is a non-issue. Maybe webpack and rollup will handle this correctly. From the docs this seems to be more pointing towards commonjs compatibility (which we wouldn't need to worry about being the commonjs build is preserved). I think we'll just have to make sure we test it well and it doesn't break an existing webpack or rollup build. 

> I've tried doing npm run build with fresh node_modules, works fine for me.

That's not what I meant. I've got a webpack 3 project that uses kefir. When I tried to build with that project it choked on kefir by complaining about some of my babel plugins. I'm not sure why exactly because I've set babel-loader to ignore the node_modules directory...

Regardless, I've just been unable to test it yet.

rpominov:
In our case `require('your-lib')` will equal to `require('your-lib')['default']`, because in `export {...};export default Kefir;` the `{...}` is the same as `Kefir`. So, we should be fine.

I just worry that collecting all exports in single object (`Kefir`) may ruin tree shaking. But we probably ruin it by collecting all methods in Observable prototype anyway, as @mAAdhaTTah pointed out above :(

mAAdhaTTah:
> I'm not sure why exactly because I've set babel-loader to ignore the node_modules directory...

I've had this problem. If you're symlinking the build, it uses the resolved path, so the regex that ignores node_modules doesn't apply. If you update `test` for your loader to be something like `/(node_modules|kefir)/`, it'll work how you expect.

zephraph:
Amazing hint. Thanks!

mAAdhaTTah:
@zephraph Were you able to test this? Any progress?

zephraph:
Yep! Heh, as others had said it had very, very little impact.

mAAdhaTTah:
I hate to throw away good work, but if it doesn't have a huge impact, I think not complicating the build process would be preferable here. I can be convinced, so I'll leave this open for a bit, but I'm leaning against.

zephraph:
Sorry, I haven't had time to grab the branch and set everything back up. I'm just going to close the PR for now. It's always here for historical context so if we decide to enable it later we can revisit.

mAAdhaTTah:
@zephraph Actually, I have a use case for an ESM build: I'm going to be teaching Observables to my coworkers, and I'd like to be able to set them up with an `index.html` and `app.js` file that they can just start with something like:

```js
import Kefir from 'https://some.cdn.com/kefir/dist/kefir.esm.js';
```

and get started. I think the idea of having everyone use ES6 w/ ESM without a build process would be kind of neat to see in practice. This is just for an exercise, so the fact that it could end up serving a bunch of files doesn't really matter here.

@rpominov I'm willing to accept that's not necessarily a strong enough reason to distribute an ESM build and thus complicate the build process, but thought I would throw it out there. Thoughts?

zephraph:
I'd have to ramp back up on this. All things considered I don't really see this as vastly increasing the complexity of the build. There were some concerns we had about it potentially affecting how the imports worked though. We'll likely need to revisit that topic.

mAAdhaTTah:
@zephraph Yeah, I'm inclined to agree, looking at the changes now. Mostly curious for @rpominov's opinion before I merge this.

rpominov:
Regarding the way imports work, there are 4 ways to import Kefir:

```js
import Kefir from 'kefir'
const Kefir = require('kefir')
import {Kefir} from 'kefir'
const {Kefir} = require('kefir')
```

We should make sure the first two still work of course 😅  

As for other two, I'm not sure which of them we support now, tbh. But we probably should try to preserve what we support if possible. It shouldn't be a problem if supporting tree shaking isn't our goal.

mAAdhaTTah:
@rpominov This PR doesn't change the current build, but creates a new one that's an ESM.

That said, does this:

```js
const Kefir = require('kefir')
```

actually work? I _might_ have thought it would need to be

```js
const Kefir = require('kefir').default
```

but I'm not as familiar with rollup (and I also don't really use CJS at all anymore).

rpominov:
> This PR doesn't change the current build, but creates a new one that's an ESM.

I still worry that someone could use `import {Kefir} from 'kefir'` and their setup configured to pick up ESM bundles if available, and then Kefir's bundle become available but doesn't support `import {Kefir}`, and build breaks. But maybe I shouldn't be so paranoid, seems like a pretty rare case.

> does this ... actually work?

Looking at the build, it should. We put every property that present in `Kefir` object directly at `exports` as well, which makes `exports` equivalent to Kefir`:

https://github.com/kefirjs/kefir/blob/1a1751c32d177c20702bc515dc5c41c3b08bd47e/dist/kefir.js#L3671-L3698

mAAdhaTTah:
Ahh, I see, fair enough.

By default, any bundler & node is going to use `pkg#main`, so there shouldn't be any conflict there either.

mAAdhaTTah:
Do we want to remove `pkg#module` for now?

zephraph:
That might be a good compromise in the interm. Otherwise it really is potentially a transparent breaking change. Alternatively we could just adjust the exports how we want them to be and do a major sem-ver release. So long as the change is transparent and well documented it might be worth just doing it right.

mAAdhaTTah:
I don't think this is worth a major version bump for Kefir; the exports don't really _need_ to be changed yet. We should consider how we're going to tree shake, but that requires a deeper investigation to do right.

mAAdhaTTah:
@zephraph Can you update this PR to remove the inclusion of `pkg#module` for now and squash your commits? Either that or allow maintainers to update ur branch.

@rpominov If we make that change, any lingering objections? Would love to get this in so I can finalize my lesson for Friday.

zephraph:
Maintainer edits are allowed. Github allows squash merges, right? Should just be able to use that when it's ready.

mAAdhaTTah:
> Github allows squash merges, right?

True story, forgot about that.

mAAdhaTTah:
Merged! Need to get this exercise done. @rpominov If you have any objections, we can revert.

rpominov:
Sorry, missed the comments. No objections 👍

Issue #249, opened 2017-10-02
How to resolve circular dependencies?


jeron-diovis:
I have a situation almost like [here](https://stackoverflow.com/questions/34343869/how-to-handle-circularly-dependent-observables-in-rxjs/34364261): to send request I need data from some stream, and that stream should be updated when request is completed. I.e., it's a circular dependency. 

In RxJS this problem is recommended to solve either with [intermediate Subject](https://stackoverflow.com/a/34364261/3437433) or [with `expand` operator](https://stackoverflow.com/a/34246857/3437433) (which I don't understand for now – looks just like flatMap, what's the difference?)

Kefir does not have subjects, but it has [`pool`](https://rpominov.github.io/kefir/#pool). Pool does the trick, but in general this solution looks just like [this](https://github.com/rpominov/kefir/issues/42#issuecomment-72114359): pool is subscribed to itself. The problem is that such pool after activation **always remains active**:

<img width="204" alt="screenshot 2017-10-02 12 56 16" src="https://user-images.githubusercontent.com/2756868/31072562-1bfdae04-a771-11e7-84c3-38bd7bb0e6f0.png">

Doc says that pool never ends, but there is nothing about "never deactivates".
To deactivate such pool I should either terminate it at all, or unplug it from itself. 

Both cases mean writing some code specially for this purpose – I can't just forget about this pool relying on deactivation of it's single consumer, defined somewhere far far away.

Let say I can't or don't want to write this special code.

So my questions are:

1. Is it ok in general that pool isn't deactivated in described situation? Active stream with no subscriptions looks strange.

2. Is there some "official" way of resolving circular dependencies in Kefir? 
Should I really use pool and explicitly terminate / unplug it? 
Or should I install `kefir-bus` package and implement Subject-like solution? Like this:
```js
const inputBus = KefirBus()
const request$ = inputBus
  .sampledBy(...some signal...)
  .flatMap(...api call...)
  .map(x => {
    inputBus.value(...)
    return x
  })
```
Or smth else?

mAAdhaTTah:
> Is it ok in general that pool isn't deactivated in described situation? Active stream with no subscriptions looks strange.

It might be OK if the resources that are left open as a result of the stream never getting deactivated don't cause a memory leak. That said, I feel like this is a bug, so I'll take a look at that.

> Is there some "official" way of resolving circular dependencies in Kefir?

[...]

PR #256, opened 2017-10-25
Add failing test case for #249

mAAdhaTTah:
Subscribing to a circular dep should cause everything to
deactivate but doesn't.

I don't have a solution to this yet, but if anyone has ideas why this might be happening, we at least have a test case we can use to verify the solution.

### DIFF
--- a/test/specs/pool.js
+++ b/test/specs/pool.js
@@ -80,6 +80,19 @@ describe('pool', () => {
     })
   })
 
+  it('should deactivate self in circular dependency', () => {
+    const pool = Kefir.pool()
+    pool.plug(pool.map(x => x))
+
+    const sub = pool.observe()
+
+    expect(pool).toBeActive()
+
+    sub.unsubscribe()
+
+    expect(pool).not.toBeActive()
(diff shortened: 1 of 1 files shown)


rpominov:
This seems like something impossible to fix without some kind of global supervision that would analyze all observables and detect loops. I mean from the point of view of individual observables there is no way to notice a loop and deactivate itself, i.e. notice that the only subscriber that left is effectively the same observable.

mAAdhaTTah:
> notice that the only subscriber that left is effectively the same observable.

I'd probably have to dig into the internals a bit more, but if the only subscriber isn't active, we don't necessarily need to detect a loop, do we?

rpominov:
> but if the only subscriber isn't active

A subscriber isn't an observable it's just a callback. The callback is added by an observable, but we have access only to a callback. But even if we had access to the observable it's in fact active because it also have a subscriber, (basically from the first observable, via another intermediate observable).

bpinto:
If this isn't a bug, should we consider closing this issue?

PR #257, opened 2017-10-26
Add thru and tests

### LINKED ISSUE #255: Kefir subclasses can't return same type
I have an ActionObservable that extends the built-in Kefir.Stream and adds an `ofType` method to it. When I have an instance of said ActionObservable, if I `map` over its values, it currently cannot return an instance of ActionObservable. I would return an instance of whatever I extended off of (in this case, Kefir.Stream).

It doesn't seem like there's an easy way to do this by even mucking about in the internals, so I'm wondering if this is a use-case we want to enable and if so, how would we go about doing so. Thoughts?

mAAdhaTTah:
Fixes #255.

* [x] Update docs
* [x] Update flowtypes

### DIFF
--- a/docs-src/descriptions/one-source.pug
+++ b/docs-src/descriptions/one-source.pug
@@ -829,8 +829,24 @@ pre(title='events in time').
                  12      14
 div
 
++descr-method('thru', 'thru', 'obs.thru(transformer)').
+  Calls #[b transformer] with #[b obs] as argument and returns whatever
+  #[b transformer] returned. Put another way, #[tt o.thru(fn)] is the
+  same as #[tt fn(o)]. This allows you to integrate your helper functions
+  into chains of Kefir method calls without adding them to Observable prototype.
 
+pre.javascript(title='example')
+  :escapehtml
+    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5, 6]);
+    var transformer = obs => obs.filter(x => x > 2);
+    var result = source.thru(transformer).map(x => x - 1); // is the same as transformer(source).map(...)
+    result.log();
+div
 
+pre(title='events in time').
+  source:  ---1---2---3---4---5---6X
+  result:  -----------2---3---4---5X
+div
 
 +descr-method('with-handler', 'withHandler', 'obs.withHandler(handler)').
   The most general transformation method. All other transformation methods above
--- a/kefir.js.flow
+++ b/kefir.js.flow
@@ -104,6 +104,7 @@ declare class Observable<+V,+E=*> {
   slidingWindow(max: number, min?: number): Observable<V[],E>;
   bufferWhile(predicate?: (value: V) => boolean, options?: {flushOnEnd?: boolean}): Observable<V[],E>;
   transduce(transducer: any): Observable<any,E>;
+  thru<V2>(callback: (this => V2)): V2;
   withHandler<V2,E2>(handler: (emitter: Emitter<V2,E2>, event: Event<V,E>) => void): Observable<V2,E2>;
 
   combine<V2,E2>(otherObs: Observable<V2,E2>, $?: empty): Observable<[V,V2],E|E2>;
--- a/docs-src/includes/side-menu.pug
+++ b/docs-src/includes/side-menu.pug
@@ -74,6 +74,7 @@ ul.toc-section
   +method('#buffer-with-count', 'bufferWithCount')
   +method('#buffer-with-time-or-count', 'bufferWithTimeOrCount')
   +method('#transduce', 'transduce')
+  +method('#thru', 'thru')
   +method('#with-handler', 'withHandler')
 
 a.toc-title(href='#combine-observables') Combine observables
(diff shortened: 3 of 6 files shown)


mAAdhaTTah:
Docs & types are here. Don't think the types / implementation will be controversial but definitely give the docs a read-over. I'm almost inclined to link directly to `most.js`'s docs on `thru` for more context. Thoughts?

docs-src/descriptions/one-source.pug
```diff
@@ -829,8 +829,25 @@ pre(title='events in time').
                  12      14
 div
 
++descr-method('thru', 'thru', 'obs.thru(callback)').
+  This transformation method takes a single callback and calls the callback with
+  #[b obs] observable. Based on the most.js method of the same name, this provides
```
rpominov: Maybe describe it like "... this allows to integrate your helper functions to chains of Kefir method calls without adding them to Observable prototype"? And then in the example show something like this `source.map(...).thru(callback).filter(...)`? I just think we should mention chaining use case explicitly. 

Also I'd call the argument `transformer` instead of `callback`. And maybe mention in the description that `o.thru(fn)` is equivalent to `fn(o)`, I think this should immediately click in reader's mind what that method does.

kefir.js.flow
```diff
@@ -104,6 +104,7 @@ declare class Observable<+V,+E=*> {
   slidingWindow(max: number, min?: number): Observable<V[],E>;
   bufferWhile(predicate?: (value: V) => boolean, options?: {flushOnEnd?: boolean}): Observable<V[],E>;
   transduce(transducer: any): Observable<any,E>;
+  thru<V2>(callback: (Observable<V,E> => V2)): V2;
```
rpominov: Can it be written as `thru<V2>(callback: this): V2;`?

mAAdhaTTah: Oh, probably. Good call.

mAAdhaTTah:
@rpominov Made some changes. Not entirely sure I'm satisfied w/ the docs, but let me know what you think.

docs-src/descriptions/one-source.pug
```diff
@@ -829,8 +829,26 @@ pre(title='events in time').
                  12      14
 div
 
++descr-method('thru', 'thru', 'obs.thru(callback)').
+  This transformation method takes a single callback and calls the callback with
+  #[b obs] observable. Based on the most.js method of the same name, this
+  allows you to integrate your helper functions into chains of Kefir method calls
+  without adding them to Observable prototype. Functions that take and return
+  an observable can be passed to #[b thru], allowing the fluent interface to be
+  used consistently.
```
rpominov: Maybe rewrite it like this?

```
+descr-method('thru', 'thru', 'obs.thru(transformer)').
  Calls #[b transformer] with the #[b obs] as argument and returns whatever 
  transformer returned. `o.thru(fn)` is the same as `fn(o)`. This allows you to 
  integrate your helper functions into chains of Kefir method calls without adding 
  them to Observable prototype. Based on the most.js method of the same name.
```

I.e. get to the point more quickly and remove some redundancy.

Also I'm on the fence about mentioning that it was based on most.js method. I'm totally ok with giving them credit for the idea. But as a docs reader, this makes me think that in order to fully understand this method I need to make myself familiar with how it works in most.js, which is not the case.

docs-src/descriptions/one-source.pug
```diff
@@ -829,8 +829,26 @@ pre(title='events in time').
...
+  :escapehtml
+    var source = Kefir.sequentially(100, [1, 2, 3, 4, 5, 6]);
+    var callback = obs => obs.filter(x => x > 2);
+    var result = source.thru(callback).map(x => x - 1); // is the same as callback(source).map(...)
+    result.log();
+div
 
+pre(title='events in time').
+  source:  ---1---2---3---4---5---6X
+  result:  -----------3---4---5---6X
```
rpominov: Don't forget to update the output ;)

docs-src/descriptions/one-source.pug
mAAdhaTTah: That's true; my thought initially was to actually link to the most.js API docs for the function, because they have a deeper explanation, but perhaps that's going overboard. If we're not going to do that, I'll remove the most.js reference.

mAAdhaTTah:
Tweaked the wording a little bit from your suggestion, but should be good.

mAAdhaTTah:
Landed! If anyone has thoughts on #256, we could release both of those in a minor bump.

rpominov:
#256 seems hard to fix to me. I'll release a new version with `thru`.

rpominov:
Just noticed we forgot to add `log output` http://rpominov.github.io/kefir/#thru :(

mAAdhaTTah:
@rpominov That's partially intentional, because there's no `setName` call in `thru`, but we can use the specific implementation used by the example. I'll open a new PR for that.

rpominov:
I think that's ok. These output snippets are there just to show the example output without running it.

PR #258, opened 2017-10-26
Add console output for thru example

mAAdhaTTah:


rpominov:
By the way, in case you'll want to update docs without me here is the command usually I use:

```
git checkout gh-pages && git merge master && npm run build && git add . && git commit -m 'build all' && git push && git checkout master
```

PR #259, opened 2017-10-30
Migrate test suite to mocha + chai + chai-kefir

mAAdhaTTah:
This is still a work-in-progress, but I've migrated most of the tests over to use `chai-kefir`. Wanted to get some eyes on this to make sure we're all on board with this.

Couple of things to note:

1. Right now `chai-kefir` depends on `kefir` directly. I'm going to move that to a `peerDependency`. My goal is to make the `require('kefir')` currently located in the `test-helpers` to point to the project itself, although I'm not entirely sure this is possible. The issue is that it currently installed kefir as a dep of kefir, which causes the chai plugin & the lib itself to point to different versions, making the `x instanceof Property` checks fail. I also have to lower the version we're transpiling `chai-kefir` for, as we're seeing errors on Node 4 that could be fixed through a more aggressive babel config.
2. Instead of changing every file, I re-exported the pieces I moved into `chai-kefir` through the `test-helper`, which allowed them to continue to be used the same way through all the tests. I also exported chai's `expect` method through the `test-helpers`.
3. I'm seeing test failures with `throttle`, which appear to be dropping the first event. I'm using a newer version of `lolex` in `chai-kefir`, which I'm using directly, which could be the cause of this issue. Or the older version of `lolex` had a bug that was fixed that is now exposing an issue in `throttle`. I'm still investigating this, but if anyone who was involved in the timer-related work has any ideas, that would be helpful, as I'm still trying to get my head around them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: node_js
 sudo: false
+before_script:
+  - npm link
+  - npm link kefir
 node_js:
 - '4'
 - '7'
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ before_script:
   - npm link
   - npm link kefir
 node_js:
-- '4'
-- '7'
+  - 6
+  - 8
+  - stable
 notifications:
   webhooks:
     urls:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: node_js
 sudo: false
-before_script:
-  - npm link
-  - npm link kefir
 node_js:
   - 6
   - 8
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "npm run clean && npm run build-js && npm run build-docs",
     "test": "./configs/prettier.sh check && rollup -c ./configs/rollup.dev.js && mocha && flow check",
     "test-only": "rollup -c ./configs/rollup.dev.js && mocha",
-    "test-debug": "rollup -c ./configs/rollup.dev.js && node --inspect-brk `npm bin`/mocha"
+    "test-debug": "rollup -c ./configs/rollup.dev.js && mocha --inspect-brk"
   },
   "keywords": [
     "frp",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@reactivex/rxjs": "5.3.0",
     "babel-preset-es2015-loose-rollup": "7.0.0",
     "chai": "^4.1.2",
-    "chai-kefir": "^1.2.0",
+    "chai-kefir": "^2.0.1",
     "flow-bin": "0.50.0",
     "inquirer": "0.10.1",
     "mocha": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
     "build-docs": "node configs/docs.js",
     "clean": "rm -r dist index.html || true",
     "build": "npm run clean && npm run build-js && npm run build-docs",
-    "test": "./configs/prettier.sh check && rollup -c ./configs/rollup.dev.js && jasmine-node --matchall test/specs && flow check",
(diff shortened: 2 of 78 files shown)


rpominov:
Regarding dependency if nothing else work we could expose a function that accepts `Kefir` and returns what currently `chai-kefir` exports directly. Regarding Node 4, let's just not run tests in it.

I'll take a look at `#3` and on the PR overall at the and of the week.

mAAdhaTTah:
Yeah, I'm still playing around with a few ideas. I'll bump node version testing to be 6, 8, and stable in the next commit (6 & 8 are LTS, and stable obvs).

mAAdhaTTah:
My apologies in advance for all the push/commit notifications. Testing a few different configurations to get `kefir` & `chai-kefir` to play nicely, so force-pushing a lot.

mAAdhaTTah:
So dopey workaround was to `npm link && npm link kefir`, which links kefir to the global node_modules then links it back to itself, which seems to get around this dependency problem. I've changed the versions tested as above. The timer issues on throttle are the only remaining issues. If we get that sorted, I can also add a `karma` run to the test process and finally get browser tests.

rpominov:
> So dopey workaround was to `npm link && npm link kefir`

I don't quite like that to set up the dev environment people will need to run these commands. Usually, you just do `npm install` and everything is set up. Also, I worry that this will be unstable and may break when we update packages or something...

Anyway, if you like this solution, we probably should mention in README that people need to run this command in order to run tests locally.

mAAdhaTTah:
> Anyway, if you like this solution, we probably should mention in README that people need to run this command in order to run tests locally.

Honestly, no, I don't really like it. I had a conversation w/ one of the chai maintainer who suggested as you did that it was idiomatic to export a function that takes the lib we're asserting against, so that's really the way to go. I would just have to bite the bullet and release 2.0 `chai-kefir`, which I guess isn't really a big deal but was reluctant to do at first.

Any thoughts on the test failures?

rpominov:
Haven't looked into failures yet, will do soon!

mAAdhaTTah:
I've made the change IRT to turning the default export into a Kefir-accepting function but I'd also like to include an API for generating the values sent down the Kefir stream. I included a proposal in the PR [here](https://github.com/mAAdhaTTah/chai-kefir/pull/3). Would be interested in your thoughts. I definitely want to get away from merely sending an object with an `error` or `current` property, as I can definitely see that conflicting with userland needs. Abstracting that away into functions makes that easier.

rpominov:
Regarding failings. I think this is because we treat `0` timestamp specially in `throttle`: https://github.com/rpominov/kefir/blob/5968e046a329c94145a569f43ea4b66327c67b3b/src/one-source/throttle.js#L26

And by default when `lolex` emulates time it starts at `0`. In current test utils we override default value here https://github.com/rpominov/kefir/blob/5968e046a329c94145a569f43ea4b66327c67b3b/test/test-helpers.js#L135

Seems like you don't do that in `chai-kefir`.

rpominov:
Ah, btw, about Karma. Don't know if we need it. I've used them to debug tests in Chrome, but now attaching Chrome debugger to node is so easy. And we'll probably be fine if we run tests only in node since we don't use any browser specific API.

mAAdhaTTah:
Yeah, I'm super impressed by the node debugging tools / integration.

My intention was checking cross-env compatibility, but to really get value out of that, we'd probably want to wire it to Sauce Labs or something like that. I agree, no need for it for now.

mAAdhaTTah:
Phew! Converted everything over to the new API. I think we're looking good.

rpominov:
Wow, how you manage to do so much work so fast? 😃
Thank you for all the work. Looks great to me, feel free to merge!

mAAdhaTTah:
Lots of find/replace with regex's. :)

PR #269, opened 2018-01-02
Fix "invariant position" type error reported by Flow v0.62.0

hallettj:


mAAdhaTTah:
Great. I can close #268. Thanks for the fix!

PR #274, opened 2018-03-05
Fix #273

### LINKED ISSUE #273: flatMapConcat never ends
I have some code that processes events on a stream, and chooses to perform either an asynchronous operation, or simply return the value synchronously. A much simplified version of the code can be seen below. The output values must be in the same order as the input, hence the `flatMapConcat`.

    const Kefir = require('kefir')
    
    const inputs = [1, 2, 3, 4, 5, 6, 7, 8]
    
    const asyncOperation = n => Kefir.later(100, n)
    const syncOperation = n => Kefir.constant(n)
    
    Kefir.sequentially(1, inputs)
    	.flatMapConcat(n => { 
    		if (n === 2) {
    			return asyncOperation(n)
    		} else {
    			return syncOperation(n)
    		}
    	})
    	.onValue(console.log)

 With the  code above the results are:

    1
    2
    3

and the stream never ends (i.e. `onEnd` is never called), whereas I would expect all values from `1` to `8` (and the stream to end).

If I change the synchronous operation to return `Kefir.later(0, n)`, then all values are output and the stream ends, so I was thinking this must have something to do with the different between `Stream` and `Property`.

However, all values are emitted and the stream ends if I use `flatMap` instead of `flatMapConcat`, which is why I am posting this issue.

mAAdhaTTah:


mAAdhaTTah:
I notice if I set the timeout on the `sequentially` call to `100`, the error goes away. If I manipulate the `sequentially` timeout between 20-100, a different number of values get emitted, e.g. if I set it to 20, 1, 2, 3, and 8 are emitted. It appears to be dropping the intervening values based on the delay of the `later` call.

mAAdhaTTah:
Found the problem. Looks like the hack around `Kefir.constant` in AbstractPool didn't subscribe to the next source in the list.

PR #281, opened 2018-09-01
Fix issue with flatmapped stream that ends while value is being emitted

Macil:
Currently, if you call .flatMap with a callback that returns an ended property (which triggers an optimized code path in flatMap) and during dispatch of the value the flatMapped stream is ended, then an exception is thrown ("can't read length property of null"). This bug was introduced in https://github.com/kefirjs/kefir/pull/274.

I noticed this issue when updating a project to the latest version of Kefir and one of the unit tests using Kefir failed on this.

This PR contains the fix and a newly-fixed test.

(I renamed my Github account recently. It was previously `@AgentME`.)

### DIFF
--- a/test/specs/flat-map.js
+++ b/test/specs/flat-map.js
@@ -129,6 +129,21 @@ describe('flatMap', () => {
       expect(unsubs).to.equal(1)
     })
 
+    it('should handle source ending in response to synchronous value', () => {
+      let mainEm
+      Kefir.stream(em => {
+        mainEm = em
+      })
+        .flatMap(x => Kefir.constant(x))
+        .onValue(x => {
+          if (x === 1) {
+            mainEm.end()
+          }
+        })
+
+      mainEm.value(1)
+    })
+
     it('should be possible to add same obs twice on activation', () => {
       const b = send(prop(), [value(1)])
       const a = Kefir.stream(em => {
--- a/test/specs/flat-map.js
+++ b/test/specs/flat-map.js
@@ -129,19 +129,21 @@ describe('flatMap', () => {
       expect(unsubs).to.equal(1)
     })
 
-    it('should handle source ending in response to synchronous value', () => {
-      let mainEm
-      Kefir.stream(em => {
-        mainEm = em
-      })
+    it('should not error when source ends in response to synchronous value', () => {
+      let didEnd = false
+      const src = stream()
+      src
         .flatMap(x => Kefir.constant(x))
         .onValue(x => {
           if (x === 1) {
-            mainEm.end()
+            send(src, [end()])
           }
         })
-
-      mainEm.value(1)
+        .onEnd(() => {
+          didEnd = true
+        })
+      send(src, [value(1)])
+      expect(didEnd).to.be.true
     })
 
     it('should be possible to add same obs twice on activation', () => {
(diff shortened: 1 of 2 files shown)


test/specs/flat-map.js
```diff
@@ -129,6 +129,21 @@ describe('flatMap', () => {
...
+      })
+        .flatMap(x => Kefir.constant(x))
+        .onValue(x => {
+          if (x === 1) {
+            mainEm.end()
+          }
+        })
+
+      mainEm.value(1)
+    })
```
mAAdhaTTah: This test doesn't assert anything.

mAAdhaTTah:
This seems fine otherwise but the test itself doesn't validate the behavior. I think you just need to assert the created stream itself is ended.

Macil:
Yeah that's fair. I'll add an assert or two that verifies everything went through as expected. (The test is checking for a lack of exceptions; pre-patch the test will fail with an exception.)

Macil:
I updated the test to check that the stream did in fact end, and made use of some pre-existing helpers.

test/specs/flat-map.js
```diff
@@ -129,6 +129,23 @@ describe('flatMap', () => {
...
+        .onValue(x => {
+          if (x === 1) {
+            send(src, [end()])
+          }
+        })
+        .onEnd(() => {
+          didEnd = true
+        })
+      send(src, [value(1)])
+      expect(didEnd).to.be.true
```
mAAdhaTTah: Instead of the `didEnd` setup, we should probably use `expect(src).to.emit([])` assert here. This will always verify what values are emitted by the `flatMap`'ed steam. My wonder / worry is the initial value never gets emitted.

I also don't think this test activates the stream, so it could pass without actually testing the code. You can look at `chai-kefir`'s assertions to help. Otherwise, I can take a quick look at this.

mAAdhaTTah: Oh wait, `onValue` activates the stream... This mimics a circular dependency, so maybe this test is fine?

Macil: I tried making the test like this:

```js
      const src = stream()
      const result = src
        .flatMap(x => Kefir.constant(x))
        .onValue(x => {
          if (x === 1) {
            send(src, [end()])
          }
        })
      expect(result).to.emit([end()], () => {
        send(src, [value(1)])
      })
```

but I didn't really like that this version of the test is also testing that the `value(1)` doesn't make it to the expect listener at the end. I don't mind that that happens, but I don't want the test to be dependent on that fact as it distracts from the real purpose of the test (and also if we ever change that behavior, I don't want this test to stand in the way). So using .onValue() and expect().to.emit() together doesn't seem to work out very well, and I don't think there's an unconvoluted way to accomplish this test with just expect().to.emit() instead of .onValue().

mAAdhaTTah: > but I didn't really like that this version of the test is also testing that the `value(1)` doesn't make it to the expect listener at the end.

I kind of think it should. It makes the expectation clearer around what should happen when we have a circular dependency like this, so I think we should validate the entirety of what we expect to emit when a flatMapped stream ends "prematurely."

Macil: That makes sense. I've swapped the code to what's in my example.

mAAdhaTTah:
LGTM!

PR #292, opened 2019-01-22
Better atomic emitting in combine operator, check new tests

HitoriSensei:
In a scenario, when you are combining a stream from X synchronous streams that come from single source (eg. mapped), single emit in origin stream caused combined stream to emit X events (while still in dispatch loop) instead of one with final (atomic) result.

example:
        const source = stream()
        const b = a.map(x => x + 2)
        const c = a.map(x => x * 2)
        const combined = Kefir.combine([b, c])

if source emits Value(1), Value(2), then combined should emit Value([3,2]), Value([5,4]) instead of Value([3,2]), Value([5,2]), Value([5,4])

Value([5,2]) is an incorrect intermediate value

Check the updated tests for the example above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kefir",
-  "version": "3.8.7",
+  "version": "3.8.6",
   "description": "Reactive Programming library for JavaScript inspired by Bacon.js and RxJS with focus on high performance and low memory usage",
   "main": "dist/kefir.js",
   "module": "dist/kefir.esm.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kefir",
-  "version": "3.8.6",
+  "version": "3.8.7",
   "description": "Reactive Programming library for JavaScript inspired by Bacon.js and RxJS with focus on high performance and low memory usage",
   "main": "dist/kefir.js",
   "module": "dist/kefir.esm.js",
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ test/in-browser/spec/KefirSpecs.js
 index.html
 test.html
 test.js
+.idea
--- a/src/batching-queue.js
+++ b/src/batching-queue.js
@@ -1,6 +1,6 @@
 export default function BatchingQueue() {
   this.lockCounter = 0
-  this.batchingQueue = []
+  this.batchingQueue = undefined
 
   this.lock = function() {
     this.lockCounter++
@@ -9,16 +9,20 @@ export default function BatchingQueue() {
   this.release = function() {
     this.lockCounter--
 
-    if (this.lockCounter === 0 && this.batchingQueue.length > 0) {
+    if (this.lockCounter === 0 && this.batchingQueue) {
       this.flushBatchingQueue()
     }
   }
 
   this.push = function(node) {
     if (!this.lockCounter) {
-      batchedNode._emitQueued()
+      node._emitQueued()
+    } else {
+      if (!this.batchingQueue) {
+        this.batchingQueue = []
+      }
+      this.batchingQueue.push(node)
     }
-    this.batchingQueue.push(node)
   }
 
   this.flushBatchingQueue = function() {
--- a/src/batching-queue.js
+++ b/src/batching-queue.js
@@ -1,6 +1,6 @@
 export default function BatchingQueue() {
   this.lockCounter = 0
-  this.batchingQueue = undefined
+  this.batched = undefined
 
   this.lock = function() {
     this.lockCounter++
@@ -9,7 +9,7 @@ export default function BatchingQueue() {
   this.release = function() {
     this.lockCounter--
 
-    if (this.lockCounter === 0 && this.batchingQueue) {
+    if (this.lockCounter === 0 && this.batched) {
       this.flushBatchingQueue()
     }
   }
@@ -18,17 +18,17 @@ export default function BatchingQueue() {
     if (!this.lockCounter) {
       node._emitQueued()
     } else {
-      if (this.batchingQueue) {
-        this.batchingQueue.push(node);
+      if (this.batched) {
+        this.batched.push(node)
       } else {
-        this.batchingQueue = [node];
+        this.batched = [node]
       }
     }
   }
 
   this.flushBatchingQueue = function() {
     let batchedNode
-    while ((batchedNode = this.batchingQueue.shift())) {
+    while ((batchedNode = this.batched.shift())) {
       batchedNode._emitQueued()
     }
   }
--- a/src/batching-queue.js
+++ b/src/batching-queue.js
@@ -1,0 +1,30 @@
+export default function BatchingQueue() {
+  this.lockCounter = 0
+  this.batchingQueue = []
+
+  this.lock = function() {
+    this.lockCounter++
+  }
+
+  this.release = function() {
+    this.lockCounter--
+
+    if (this.lockCounter === 0 && this.batchingQueue.length > 0) {
+      this.flushBatchingQueue()
+    }
+  }
+
+  this.push = function(node) {
+    if (!this.lockCounter) {
+      batchedNode._emitQueued()
+    }
+    this.batchingQueue.push(node)
+  }
+
+  this.flushBatchingQueue = function() {
+    let batchedNode
+    while ((batchedNode = this.batchingQueue.shift())) {
+      batchedNode._emitQueued()
+    }
+  }
+}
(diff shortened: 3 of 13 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kefir",
-  "version": "3.8.6",
+  "version": "3.8.7",
```
mAAdhaTTah: Don't increment this here. We'll do this when we release.

mAAdhaTTah:
I can't say I'm in love with this solution. I'd also venture to say this is a breaking change, although the fact that our tests don't document it as such is a little worrying.

HitoriSensei:
I can't say that I'm in love with the solution either, especially the
global loop counter part, but it's the best I could do with the minimal
code changes (without rewriting the dispatching system).

I guess it could be the breaking change if there are systems that actually
depend on this behaviour. I'm actually curious how combining behaves in
other reactive frameworks.

I thought of it as a bug because I've noticed the test that handles atomic
updates, but only tested passive sources, not the active ones. And because
I actually encountered bugs in my application because of the non-atomic
behaviour I didn't suspect.

Maybe we should introduce something new like "combineAtomic" so it wouldn't
break any apps that currently use kefir?

wt., 22.01.2019, 17:47: James DiGioia <notifications@github.com> napisał(a):

Macil:
Something I'm concerned about: imagine a user has a function foo() which accomplishes some task synchronously. When foo() returns, the task is expected to have been completed. The task could be mutating some object, or even just logging to the console. Internally as an implementation detail, foo() accomplishes this task by calling a method on a Kefir stream emitter, and then a chain of listeners of that stream receive the value and do the task.

Now imagine that foo() is called by some listener of a Kefir stream "bar". (Maybe it's called in an bar.onValue callback, or even a bar.map callback.) (The bar stream and the stream used inside foo may be related.) Could this patch possibly break foo(), by causing the value emitted inside of foo to be buffered until after the event on bar has finished dispatching?

Simple unaffected example of this scenario:

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');

const innerFooBus = kefirBus();
const innerFooStream = innerFooBus; // Some Kefir observable based on innerFooBus

innerFooStream.onValue(x => {
  console.log('foo', x);
});

function foo() {
  innerFooBus.value(123);
}

const bar = Kefir.constant(12); // Any Kefir observable

bar.onValue(() => {
  console.log('pre-foo');
  foo();
  console.log('post-foo');
});

// Expected output:
// pre-foo
// foo 123
// post-foo
```

Now here's a modified version of that example which breaks when using this PR's code:

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');

const innerFooBus = kefirBus();
const innerFooStream = Kefir.combine([innerFooBus], x => x[0]); // Some Kefir observable based on innerFooBus

innerFooStream.filter(x => x === 123).onValue(x => {
  console.log('foo', x);
});

function foo() {
  innerFooBus.value(123);
}

const bar = innerFooStream.filter(x => x === 1);

bar.onValue(x => {
  console.log('pre-foo');
  foo();
  console.log('post-foo');
});

innerFooBus.value(1);

// Expected output:
// pre-foo
// foo 123
// post-foo

// Actual output with this PR:
// pre-foo
// post-foo
// foo 123
```

The example appears a bit contrived, but I believe a few large projects I've been a part of that have used Kefir have structures like this (it's easy to happen when you have a mostly OOP-style codebase that sprinkles in a lot of Kefir for dispatching messages between objects). We previously used Bacon.js and spent a lot of time attempting to debug mysterious behavior that I now believe were caused in part by this same issue existing in Bacon.js.

Is there a way to change the PR to fix this example? (Or would it make sense to add some new APIs, like `Kefir.batchingCombine`, so existing code is unaffected?)

HitoriSensei:
I think that even if your case could be solved in the PR we couldn't be
sure if there isn't any other breaking scenario.

I will update the PR to incorporate two changes:

* Introduce new "batchingCombine" helper without changing the current
"combine" to not break any projects
* Make the change more OOP friendly
* Update the tests to document currently undocumented combine case

Then we can reason again, what do you think?

śr., 23.01.2019, 02:56: Chris Cowan <notifications@github.com> napisał(a):

HitoriSensei:
I've updated the PR with the changes from my previous comment

HitoriSensei:
Hello again,

I've updated the PR branch to fix the scenario provided by @Macil.

Also, batching queue is now instanced for every "real" source and combined steam.
By "real" sources I mean streams that are not created by mapping other streams.
Streams that have only one source share it's batching queue.

I've also added provided the scenario to the test suite for combine and combineBatched.

Macil:
I'm decreasingly sure that this belongs directly in Kefir because of three things: this infects every Observable instance (possibly with a performance or memory hit), there's a lot of arbitrariness around how batching should work (that could add a lot of API surface to Kefir if embraced), and atomic updates can be implemented externally to Kefir (without monkey-patching).

<details>
  <summary>Elaboration:</summary>

---

I originally typed the following out over the last few days, and it lead me to the above opinion.

---

If a user wants to call methods on several different stream emitters and have the updates be batched to a combineBatched stream, I think there should be a supported way to do this.

I noticed that this code works with your PR:

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');

const bus1 = kefirBus();
const bus2 = kefirBus();

Kefir.combineBatched([bus1, bus2]).log('batch');

bus1.value(123);
bus2.value(456);

Kefir.merge([Kefir.constant(null)]).onValue(() => {
  bus1.value(789);
  bus2.value('abc');
});

// batch <value> [ 123, 456 ]
// batch <value> [ 789, 'abc' ]
```

Though this doesn't batch (at least in your earlier code):

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');

const bus1 = kefirBus();
const bus2 = kefirBus();

Kefir.combineBatched([bus1, bus2]).log('batch');

bus1.value(123);
bus2.value(456);

Kefir.constant(null).onValue(() => {
  bus1.value(789);
  bus2.value('abc');
});

// batch <value> [ 123, 456 ]
// batch <value> [ 789, 456 ]
// batch <value> [ 789, 'abc' ]
```

I definitely wouldn't want to document the first case as a recommended technique for batching. If the second case worked, then it might be good enough to include in the documentation as a technique for this case.

---

We could instead add a separate `Kefir.batching` API specifically for this:

```js
Kefir.batching(() => {
  bus1.value(789);
  bus2.value('abc');
});
```

Though it's for such a specific situation, so I don't really prefer this route.

---

Thinking about how there could be a separate batching API made me realize that this feature could be implemented completely externally to Kefir with an API like this:

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');
const {combine, batchHandler} = require('kefir-batching');

const bus = kefirBus();
const source = bus.withHandler(batchHandler);
const a = source.map(x => x*10);
const b = source.map(x => x*100);

const combined = combine([a, b, bus2]);
combined.log('combined');

bus.value(1);
bus.value(2);
bus.value(3);

// combined [10, 100]
// combined [20, 200]
// combined [30, 300]
```

This technique does have to be opted into here with the `.withHandler(batchHandler)` call, but having explicit batching rather than implicit batching as in this PR makes it natural to give the user more power over batching:

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');
const {combine, doBatch} = require('kefir-batching');

const bus1 = kefirBus();
const bus2 = kefirBus();

combine([bus1, bus2]).log('batch');

bus1.value(123);
bus2.value(456);

doBatch(() => {
  bus1.value(789);
  bus2.value('abc');
});

// batch <value> [ 123, 456 ]
// batch <value> [ 789, 'abc' ]
```

The kefir-batching module here could even allow non-synchronous batching over a time period. This could be especially useful in a game or framework where you don't want to .combine more than once per frame.

```js
const Kefir = require('kefir');
const kefirBus = require('kefir-bus');
const {combine, startBatch, endBatch} = require('kefir-batching');

const bus1 = kefirBus();
const bus2 = kefirBus();

combine([bus1, bus2]).log('batch');

bus1.value(123);
bus2.value(456);

startBatch();
bus1.value(789);
setTimeout(() => {
  bus2.value('abc');
  endBatch();  
}, 100);

// batch <value> [ 123, 456 ]
// batch <value> [ 789, 'abc' ]
```

Someone may even want a version of combine that auto-batches based on microtasks: when it receives a value, if a batch isn't started, then it runs `startBatch(); Promise.resolve().then(endBatch)`.

The above two possibilities could be made awkward if Kefir had its own built-in combine batching strategy.

</details>

---

I'm planning on implementing the kefir-batching module as used in the examples above.

PR #293, opened 2019-06-08
Update (dev) dependencies, address github security alert

Macil:
It seems like Github emailed a "security alert" to the maintainers of Kefir because of a dev-only subdependency. I'm pretty confident it doesn't really affect us or users of Kefir (it only matters if untrusted content gets passed to a subdependency that's only used during build), but it would be nice to see that alert go away, and it's also generally nice to have our dependencies up-to-date.

This PR is best viewed commit-by-commit. Changes to package-lock.json were done by npm (after any edits to package.json, I would remove package-lock.json and run `npm install` so npm regenerated it from scratch).

* 83493ff I updated most dev dependencies. I switched most dev dependencies to being specified by major version (using `^`) since they're pinned anyway by package-lock.json. I did not update prettier yet because that would require other files to be re-formatted, and I didn't update rollup because I'm not very familiar with it and I think updating it might mean updating to Babel 7 which is a bigger change than I want to do here. I did update Rxjs to version 6 which required a tiny change in one test. I believe this PR fixes the security alert by updating mocha and inquirer, which I think had the vulnerable subdependencies.
* 0bc3df6 I changed symbol-observable from being listed as a regular dependency to being listed as a devDependency. When we build Kefir, we build symbol-observable directly into the Kefir bundle, so there's no need for the consumers of Kefir to install symbol-observable.
* 75198fb I updated Prettier to the latest version, and let it reformat all files.
* 793474c I moved the Prettier config out of prettier.sh into Prettier's .prettierrc config file. (Presumably editor plugins will be able to make use of that too.) I also fixed an issue in prettier.sh where it would fail to run if the path to the prettier executable had spaces.
* 096d628 Updated the sinon dev dependency. I was worried this might require some changes in the tests so I didn't update this originally, but it turned out fine.

I diffed the built Kefir bundle after the first two commits (before I updated Prettier) to master. The only changes were minor ones caused by symbol-observable being updated.

This PR is mostly for internal benefits. The compiled output only changes in a minor way because of formatting changes and minor changes in symbol-observable (that I don't think change its behavior in a way important for Kefir users), so I don't think this PR alone warrants a new version being published to npm yet.


mAAdhaTTah:
Should we be bundling symbol-observable? Seems like that should be external, altho that might be a breaking change 😕.

Also, after we merge this, how do we feel about adding dependabot to this repo?

configs/prettier.sh
```diff
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-PRETTIER="`npm bin`/prettier --no-semi --no-bracket-spacing --print-width 120 --single-quote --trailing-comma es5"
+PRETTIER="`npm bin`/prettier"
```
mAAdhaTTah: We can use `npx prettier` here now.

rpominov: Yeah, originally I've created a shell script because didn't find a better way to safe the config. Perhaps we can get rid of it now.

mAAdhaTTah: Oh! Yeah, we could definitely move this into `pkg.json#scripts` entirely.

test/specs/log.js
```diff
@@ -15,18 +15,24 @@ describe('log', () => {
 
   describe('removing', () => {
     it('should return the stream', () => {
-      expect(stream().log().offLog()).to.be.observable.stream()
+      expect(
+        stream()
+          .log()
+          .offLog()
+      ).to.be.observable.stream()
```
mAAdhaTTah: There are a couple of changes like this that not only seem worse than what we had before but are reformatting lines that don't exceed the 120 char `printWidth` config. Are we sure this is configured properly?

mAAdhaTTah: It's possible this was a prettier opinion change tho.

rpominov:
> Should we be bundling symbol-observable? Seems like that should be external, altho that might be a breaking change 😕.

In my opinion it's not important, either way is good.

> Also, after we merge this, how do we feel about adding dependabot to this repo?

Our dependencies are mostly `dev`. So it's not as important to keep them up to date. 

Also, when upgrading a regular dependency you can automatically verify it using tests. This will not work for dev. So someone will have to manually check all those PRs.

test/specs/log.js
mAAdhaTTah: Yeah, sorry, testing on the prettier playground suggests this is intended.

mAAdhaTTah:
> In my opinion it's not important, either way is good.

Ok, let's leave it for now, I'll file an issue to revisit this in 4.0 if / when we decide to do that.

configs/prettier.sh
Macil: I've done that now. (Since it's just in `pkg.json#scripts` now, there's no need for npx since the scripts are run with the npm bin directory in path.)

Macil:
Thanks for the comments and making the issue about symbol-observable. I inlined the prettier stuff.

Issue #302, opened 2020-05-16
flatMapConcat should work when a spawned observable ends synchronously on activation


nashspence:
In our code, we encountered a case where an observable spawned by flatMapConcat was ending synchronously on activation while other observables were queued up.  When this happened, the next observable in the queue never activated after the first one ended.

We traced this problem to a missing edge case in abstract pool and will be submitting a pull request including a fix and corresponding test.

PR #303, opened 2020-05-16
pull request to fix issue #302

nashspence:
#302

### DIFF
--- a/src/many-sources/abstract-pool.js
+++ b/src/many-sources/abstract-pool.js
@@ -97,6 +97,12 @@ inherit(AbstractPool, Stream, {
         if (this._active) {
           this._subToEnd(obs)
         }
+      } else {
+        if (this._queue.length !== 0) {
+          this._pullQueue()
+        } else if (this._curSources.length === 0) {
+          this._onEmpty()
+        }
       }
     } else {
       this._curSources = concat(this._curSources, [obs])
--- a/test/specs/flat-map-concat.js
+++ b/test/specs/flat-map-concat.js
@@ -108,6 +108,28 @@ describe('flatMapConcat', () => {
     ])
   })
 
+  it('should work when a spawned observable ends synchronously on activation', () => {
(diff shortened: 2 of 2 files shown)


mAAdhaTTah:
Wow, interesting issue. Surprised this hasn't come up before.

The test failure is because the test isn't formatted with prettier. You can run `npm run prettify`. I otherwise ran the tests locally and everything looks good. Thanks for the contribution!

nashspence:
No worries. Yah, we were a bit surprised too but there you have it. I’ll check into the formatting issue, I thought I had prettier’ed it but I must have forgotten.

nashspence:
In running the npm script for prettier I get a lot of changes unrelated to this issue. Would you like those included? Otherwise, I’m seeing no changes on the files this touches.

mAAdhaTTah:
I wonder if it's a prettier version mismatch. It kinda looks similar on the Travis output (other files are failing but not the one you changed), but when I run it locally, I only get a change in your file. Two questions:

1. What's your node & npm version?
2. Did you install with `npm install` or `npm ci`?

nashspence:
Could for sure be something like that. I’ll try reinstalling and testing it again. Node v13.13.0

nashspence:
Yay, totally my bad on that one. I had included my fork in my mono-repo as a submodule when figuring out the issue, which does dependency hoisting and whatnot. After isolating it and using the normal npm install procedure I’m seeing what you’re seeing and all is well. Will update the PR in a second here

nashspence:
Hm, still failing I see.

mAAdhaTTah:
Only Node 6. My guess is the older npm version installs a newer version of prettier. I'm going to merge & fix this in a separate PR.

Thank you for the contribution!

nashspence:
No worries. Thanks for what you do as well!

mAAdhaTTah:
This has been published as v3.8.7

PR #309, opened 2020-12-02
Track active observables

bpinto:
Hey all, first of all thanks for creating and managing this library.

I'd like to discuss adding support for tracking active observables so that we may support features like this on tests:

```
import { activeObservables, clearActiveObservables } from 'kefir'

beforeEach(() => { clearActiveObservables() })
afterEach(() => { expect(activeObservables.length).toEqual(0) })
```

and to eventually create a browser extension or even simpler to do a `$K.activeObservables` on the browser console - assuming you've bound kefir to window `window.$K = Kefir` - and see if there are any memory leaks on your code.

I've pushed a few commits here that would not be part of the final PR:

**1. dockerize**

I use docker for everything so I don't need to install `npm` and alike on my computer, so I created simple files to work on kefir using docker. Not sure if it's useful for anyone, but it's unrelated to this change.

**2. [Remove me]** Commit that has this label are to be removed and I added it here to confirm the generated files (`dist/*`) are not affected and only the new `kefir.dev.js` file is changed by this code.

---

I've opened a PR instead of an issue, to show some ideas, but this is meant to be a discussion more than a mergeable PR at this point.

I'm not confident on my [Proxy](https://javascript.info/proxy) usage. Perhaps someone has another idea on how to track observables?

If the idea is interesting, would it make sense to change our tests to check if all observables have been unsubscribed?


mAAdhaTTah:
Definitely going to need to ponder this a bit.

> I'm not confident on my Proxy usage. Perhaps someone has another idea on how to track observables?

I don't think we need to go full Proxy. You can just monkeypatch the method. If you wanted to use Proxy, you should probably Proxy the entire Observable object and then trap the particular methods in the Proxy handlers.

bpinto:
> You can just monkeypatch the method.

What is your suggestion so I could try? One problem I had before was that `Kefir.atom` and other will extend this Observable class and thus remove any implementation we set on `_onActivation` and `_onDeactivation`.

mAAdhaTTah:
> What is your suggestion so I could try implementing it?

Ultimately not terribly different from what you have, except instead of returning a proxy, you return a new function.

> One problem I had before was that Kefir.atom and other classes will extend this Observable class and thus remove any implementation we set on `_onActivation` and `_onDeactivation`.

Not 100% sure I follow; if you have an example, that would be helpful. But if we're wrapping the method, because we still end up calling the underlying method, I don't think it'll be an issue.

bpinto:
> Ultimately not terribly different from what you have, except instead of returning a proxy, you return a new function.

Ok, I'll update this PR soon.

> Not 100% sure I follow; if you have an example, that would be helpful. But if we're wrapping the method, because we still end up calling the underlying method, I don't think it'll be an issue.

I misunderstood what you said earlier, I thought you had suggested we wouldn't call the original method. 👍

bpinto:
Okay, so I've pushed another code which does not use Proxy, now there is something weird happening, the following test would not unsubscribe the observable:

```
export const getValue = maybeObservable => {
  if (!isObservable(maybeObservable)) return maybeObservable
  let value
  const setValue = val => value = val
  const firstValue = maybeObservable.take(1)
  firstValue.onValue(setValue)
  firstValue.offValue(setValue)
  return value
}

test('returns syncronous value from observables', () => {
  expect(streams.getValue(K.constant(1))).toBe(1)
})
```

The following observable is still tracked by `Kefir.activeObservables`:

```
    [ AnonymousObservable {
        _dispatcher: null,
        _active: false,
        _alive: false,
        _activating: false,
        _logHandlers: null,
        _spyHandlers: null,
        _onActivation: [Function: _onActivation],
        _onDeactivation: [Function: _onDeactivation],
        _currentEvent: { type: 'value', value: 1 },
        _source: null,
        _name: 'constant.take',
        _n: 0,
        '_$handleAny': null } ]
```

I don't understand yet why this observable (which is not active, and not alive) has not been removed from the array, but I'll debug it further.

P.S.: Would you rather use `apply` from [src/utils/function](https://github.com/kefirjs/kefir/blob/59d7aa7c45f026ec9761f34c595f35ba5aaccd39/src/utils/functions.js#L30) ?

bpinto:
@mAAdhaTTah I've finally come back to this PR and while there are commits that I need (some I have already)  to extract into smaller PRs, this is ready for review.

I have a [commit](https://github.com/kefirjs/kefir/pull/309/commits/6018832118ef12607e63a9ab1f12008c8737861d) where I had to brute-force subscription deactivation because I don't think there is another way to handle it with the existing code. I'm not sure if it's worth exporting the onAny function on `toPromise` or if the subscription is deactivating automatically on non-test environments because a promise always ends.

bpinto:
@mAAdhaTTah I'm thinking that maybe we shouldn't introduce a new dev dist and just include these methods on the normal distributed files?

mAAdhaTTah:
Maybe I should ask more generally: what's the purpose of this? Why do you need to track active observables?

bpinto:
For now, we are only tracking active observables to track memory leaks.

```
afterAll(() => {
  if (activeObservables.length !== 0) {
    fail()
  }
})

test('all observables are unsubscribed', () => {
  render(<MyComponent />)
})
```

It's very easy to make a mistake and forget to unsubscribe a component when using `.observe` for instance.

bpinto:
We could, though I think that tracking observables during a test environment is not the only usage. We could for instance want to run this on our browser window and make sure that while navigating from one page to another the number of observables is corresponding to what we expect.

I was thinking about using a configurable environment variable that would default to the test environment so that users could opt in and out in any environment.

```
const { NODE_ENV, TRACK_OBSERVABLES }
const trackObservables = TRACK_OBSERVABLES || NODE_ENV === 'test'
```

PR #315, opened 2021-05-12
Adds optional callback method (logger) for Observable.log/spy()

unematiii:
I have a use-case where I would like to log/spy using function other than `console.log` and I don't want to override the `log` method on `console`. This PR should make this possible. 

(Ofc, it is possible to simply wrap the observer and build the desired functionality yourself, but it would be more convenient, if this option comes from library itself.)


bpinto:
I personally like this change. On https://github.com/kefirjs/kefir/pull/318 I'm proposing we drop support for node 6 and 8 (and maybe 10) which means tests would pass.

PR #317, opened 2021-08-14
Fix most audit warnings

bpinto:
- [x] npm install: `version` attribute change on `package-lock.json`.
- [x] npm audit fix.

There are 5 remaining reported vulnerabilities to be fixed, these will be done separately as they might require code changes.


Macil:
What version of npm are you using? With or without this PR, while using the latest npm 7.20.6, I get a lot of changes to the package-lock.json after running `npm install`. Could you make sure you're using the latest npm (`npm i -g npm`), and then entirely regenerate the package-lock.json (delete package-lock.json and node_modules and then run `npm install` again)?

bpinto:
> What version of npm are you using?

`6.14.14`

> With or without this PR, while using the latest npm 7.20.6, I get a lot of changes to the package-lock.json after running npm install. Could you make sure you're using the latest npm (npm i -g npm), and then entirely regenerate the package-lock.json (delete package-lock.json and node_modules and then run npm install again) so the package-lock exists in the latest format

I'll do that. 👍 I don't think there is a way to specify a `npm` version to be used on a project, is there?

> pins the latest compatible version of everything

Would you like me to pin it using a fixed version or `~` or `^` ?

Macil:
>Would you like me to pin it using a fixed version or `~` or `^`?

It shouldn't be necessary to manually update the package.json. Just running npm install without a package-lock.json will create a new one using the latest versions that are compatible with the ranges already present in package.json.

Macil (approved):

PR #318, opened 2021-08-14
Mocha upgrade

bpinto:
This is a controversial PR since it removes automated tests for no longer maintained nodes (6, 8, 10) as mocha no longer supports these node versions.

We may want to upgrade mocha to a not so recent version as for instance, mocha v7 only removed support for node 6.


bpinto:
I thought I was going to need a newer version of mocha for #309 but I don't think that's the case as I didn't end up using the [Root Hook plugins](https://mochajs.org/#root-hook-plugins) approach - it didn't work as the plugin files are only loaded once and `test-helper` overrides Kefir on watch-mode.

In any case, I think dropping official support node 6 and 8 is probably something we want to do so that we can keep the dependencies up to date. While I think dropping support for node 10 might be needed to get the latest version of some libraries, we may want to hold it a little bit.

But, please let me know what you think and I'll do the changes (or close this PR).

bpinto:
Any opinions about this? I'd like to upgrade other dev dependencies though some of them no longer work on some node versions, so the decision we take here will have an impact on which other dev dependencies we could upgrade.

It's worth mentioning that the distributed code is not incompatible with those older node versions, it's just that we wouldn't be able to run the tests there.

rpominov:
Thanks for the PR. Was easier for me to do it in master as I fixing other things.

PR #319, opened 2021-08-15
npm run watch

bpinto:
Automatically run tests when src or test file changes.


bpinto:
I'm going to close this since it's not a good idea with rollup. I need to learn more about mocha before I can do this.

PR #320, opened 2021-08-19
Add overlapping option to flatMap

bpinto:
This will cause the next stream to be added before any old streams are removed.

```
const channelA = Kefir.stream((emitter) => {
  console.log('connect a');
  let count = 0, id = setInterval(() => emitter.value(count++), 250);
  return () => { console.log('disconnect a'); clearInterval(id); };
});

const channelB = Kefir.stream((emitter) => {
  console.log('connect b');
  let count = 0, id = setInterval(() => emitter.value(count++), 250);
  return () => { console.log('disconnect b'); clearInterval(id); };
});

const data = {
  a: channelA,
  b: Kefir.combine([channelA, channelB]),
  c: channelB,
};

Kefir.sequentially(1000, ['a', 'b', 'c', undefined])
  .flatMapLatest(p => p ? data[p] : Kefir.never())
  .log('result');
```

With overlapping option disabled (default):

```
> connect a
> result <value> 0
> result <value> 1
> result <value> 2
> disconnect a
> connect a
> connect b
> result <value> [0, 0]
> result <value> [1, 0]
> result <value> [1, 1]
> result <value> [2, 1]
> result <value> [2, 2]
> disconnect a
> disconnect b
> connect b
> result <value> 0
> result <value> 1
> result <value> 2
> disconnect b
> result <end>
```

With overlapping option enabled:

```
> connect a
> result <value> 0
> result <value> 1
> result <value> 2
> connect b
> result <value> [3, 0]
> result <value> [3, 1]
> result <value> [4, 1]
> result <value> [4, 2]
> disconnect a
> result <value> 3
> result <value> 4
> result <value> 5
> disconnect b
> result <end>
```

Closes: #235 #236

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -327,8 +327,10 @@ import FlatMap from './many-sources/flat-map'
 Observable.prototype.flatMap = function(fn) {
   return new FlatMap(this, fn).setName(this, 'flatMap')
 }
-Observable.prototype.flatMapLatest = function(fn) {
-  return new FlatMap(this, fn, {concurLim: 1, drop: 'old'}).setName(this, 'flatMapLatest')
+Observable.prototype.flatMapLatest = function(fn, options = {}) {
+  options.concurLim = 1
+  options.drop = 'old'
+  return new FlatMap(this, fn, options).setName(this, 'flatMapLatest')
 }
 Observable.prototype.flatMapFirst = function(fn) {
   return new FlatMap(this, fn, {concurLim: 1}).setName(this, 'flatMapFirst')
--- a/src/many-sources/abstract-pool.js
+++ b/src/many-sources/abstract-pool.js
@@ -5,12 +5,13 @@ import {concat, forEach, findByPred, find, remove, cloneArray} from '../utils/co
 
 const id = x => x
 
-function AbstractPool({queueLim = 0, concurLim = -1, drop = 'new'} = {}) {
+function AbstractPool({queueLim = 0, concurLim = -1, drop = 'new', overlapping = false} = {}) {
   Stream.call(this)
 
   this._queueLim = queueLim < 0 ? -1 : queueLim
   this._concurLim = concurLim < 0 ? -1 : concurLim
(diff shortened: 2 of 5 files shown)


bpinto:
This is my first time diving into kefir codebase so I most likely did something silly while trying to implement this feature. In any case, please let me know what you think!

src/index.js
```diff
@@ -327,8 +327,10 @@ import FlatMap from './many-sources/flat-map'
 Observable.prototype.flatMap = function(fn) {
   return new FlatMap(this, fn).setName(this, 'flatMap')
 }
-Observable.prototype.flatMapLatest = function(fn) {
-  return new FlatMap(this, fn, {concurLim: 1, drop: 'old'}).setName(this, 'flatMapLatest')
+Observable.prototype.flatMapLatest = function(fn, options = {}) {
+  options.concurLim = 1
```
bpinto, line 331: If only we had object rest spread... 😬

test/specs/flat-map.js
```diff
@@ -239,4 +240,90 @@ describe('flatMap', () => {
       expect(result).to.flowErrors(c)
     })
   })
+
+  describe('overlapping with a concurrency limit that has maxed out', () => {
+    describe('and with a queue limit', () => {
+      it('not maxed out, should add to the queue', () => {
```
bpinto, line 246: All these "should add to the queue" tests are kinda silly because I can't really confirm that the queue has changed...

src/many-sources/abstract-pool.js
```diff
@@ -148,8 +154,8 @@ inherit(AbstractPool, Stream, {
     return index
   },
 
-  _removeCur(obs) {
-    if (this._active) {
+  _removeCur(obs, after) {
```
bpinto, line 157: I don't like passing a boolean, not sure about performance impact, but wouldn't it be better to have two methods?

`_removeCur` and `_removeCurWithDelayedUnsubscription`

